PR #2, opened 2021-11-16
Import hacks

jagerman:


majestrate:
from hax import hax

jagerman:
> from hax import hax

"I'm not exactly a 1337 h4x0r, but I do consider myself a hacker."

PR #6, opened 2021-11-19
Add many administration capabilities to `python3 -msogs`

jagerman:
Fleshes out various admin function that wasn't implemented, adds docs, more output, etc.


sogs/__init__.py
```diff
@@ -3,8 +3,3 @@
 from . import config
 from . import logging
 from . import db
-from . import routes
```
majestrate, line 6: this file is best to be totally empty but still existing. it was only there to have it runnable via `sogs:app`

majestrate (approved):

PR #7, opened 2021-11-22
pysogs zmq bot api

majestrate:
bot api for sogs

### DIFF
--- a/sogs/events.py
+++ b/sogs/events.py
@@ -33,6 +33,7 @@ def _user_from_conn(conn):
     """
     make a model.User from a connection using its curve pubkey as the session id.
     """
+    # TODO: blinding?
     return model.User(session_id='05' + conn.pubkey.hex())
 
 
@@ -99,7 +100,7 @@ def _collect_bytes(iterable: Iterable[bytes]):
 
 
 @api(name='sub', minargs=1)
-def handle_subscribe(*events, conn=None):
+def subscribe(*events, conn=None):
     """ subscribe connection to many events """
     sub = set()
     for ev in events:
@@ -115,7 +116,7 @@ def handle_subscribe(*events, conn=None):
 
 
 @api(name='unsub', minargs=1)
-def handle_unsubscribe(*events, conn=None):
+def unsubscribe(*events, conn=None):
     """ unsub connection to many events """
     unsub = set()
     for ev in events:
@@ -132,8 +133,8 @@ def handle_unsubscribe(*events, conn=None):
 
 
 @api(name="request", minargs=2)
-def handle_rpc_call(method, path, body=None, *, conn=None):
-    """ make a sub request via zmq """
+def request(method, path, body=None, *, conn=None):
+    """ make an rpc request via zmq """
     ctype = None
     # guess content type
     if body:
@@ -141,7 +142,6 @@ def handle_rpc_call(method, path, body=None, *, conn=None):
             ctype = 'application/json'
         else:
             ctype = 'application/octet-stream'
-
     resp = make_subrequest(
         method.decode('ascii'), path.decode('ascii'), content_type=ctype, body=body
     )
--- a/sogs/events.py
+++ b/sogs/events.py
@@ -1,0 +1,149 @@
+from collections import defaultdict
+
+from oxenmq import AuthLevel
+
+from . import model
+from .omq import omq
+from .web import app
+
+from binascii import hexlify
+from types import Iterable
+
+from oxenc import bt_serialize
+
+from .routes.subrequest import make_subrequest
+
+from flask import g
+
+# pools for event propagation
+_pools = defaultdict(list)
+
+status_OK = 'OK'
+status_ERR = 'ERROR'
+
+# the events we are able to subscribe to
+EVENTS = ('message', 'joined', 'parted', 'banned', 'unbanned', 'deleted', 'uploaded')
+
+
+def event_name_valid(eventname):
+    """ return True if this event name is something well formed """
+    return eventname in EVENTS
+
+
+def _user_from_conn(conn):
+    """
+    make a model.User from a connection using it's curve pubkey as the session id.
+    """
+    return model.User(session_id='05' + hexlify(conn.pubkey).decode())
+
+
+def _propagate_event(eventname, *args):
+    """ propagate an event to everyone who cares about it """
+    assert event_name_valid(eventname)
+    global omq, _pools
+    sent = 0
+    for conn in _pools[eventname]:
+        omq.send(conn, f'sogs.event.{eventname}', *(bt_serialize(a) for a in args))
+        sent += 1
+    if sent:
+        app.logger.info(f"sent {eventname} to {sent} subscribers")
+
+
+_category = omq.add_category('sogs', AuthLevel.basic)
+
+
+def api(f, *, name=None, minargs=None):
+    """ set up a request handler for zmq for a function with name of the endpoint """
+    assert name is not None
+
+    def _handle_request(msg):
+        try:
+            if minargs and len(msg.data) < minargs:
+                raise ValueError(f"Not enough arguments, got {len(msg.data)} expected 2 or more")
+            app.logger.debug(f"zmq request: {name} for {msg.conn}")
+            g.user = _user_from_conn(msg.conn)
+            retval = f(*msg.data, conn=msg.conn)
+            if retval is None:
+                msg.reply(status_OK)
+            elif isinstance(retval, tuple):
+                msg.reply(status_OK, *retval)
+            else:
+                msg.reply(status_OK, bt_serialize(retval))
+        except Exception as ex:
+            app.logger.error(f"{f.__name__} raised exception: {ex}")
+            msg.reply(status_ERR, f'{ex}')
+        finally:
+            g.user = None
+
+    global _category
+    _category.add_request_command(name, _handle_request)
+    app.logger.info(f"register zmq api handler: sogs.{name}")
+    return f
+
+
+def _collect_bytes(iterable: Iterable[bytes]):
+    """ collect all bytes from an iterable of bytes and put it into one big bytes instance """
+    data = bytes()
+    for part in iterable:
+        data += part
+    return data
+
+
+@api(name='sub', minargs=1)
+def handle_subscribe(*events, conn=None):
+    """ subscribe connection to many events """
+    sub = set()
+    for ev in events:
+        name = ev.decode('ascii')
+        if not event_name_valid(name):
+            raise Exception(f"invalid event type: {name}")
+        sub += name
+
+    global _pools
+    for name in sub:
+        _pools[name].append(conn)
+    app.logger.debug(f"sub {conn} to {len(sub)} events")
+
+
+@api(name='unsub', minargs=1)
+def handle_unsubscribe(*events, conn=None):
+    """ unsub connection to many events """
+    unsub = set()
+    for ev in events:
+        name = ev.decode('ascii')
+        if not event_name_valid(name):
+            raise Exception(f"invalid event type: {name}")
+        unsub += name
+
+    global _pools
+    for name in unsub:
+        _pools[name].remove(conn)
+    app.logger.debug(f"unsub {conn} to {len(unsub)} events")
+
+
+@api(name="request", minargs=2)
+def handle_rpc_call(method, path, body=None, *, conn=None):
+    """ make a sub request via zmq """
+    ctype = None
+    # guess content type
+    if body:
+        if body[0] in (b'{', b'['):
+            ctype = 'application/json'
+        else:
+            ctype = 'application/octet-stream'
+
+    resp = make_subrequest(
+        method.decode('ascii'), path.decode('ascii'), content_type=ctype, body=body
+    )
+    return resp.status_code, _collect_bytes(resp.response)
+
+
+class _Notify:
+    """ Holder type for all event notification functions """
+
+
+notify = _Notify()
+
+# set up event notifiers
+for ev in EVENTS:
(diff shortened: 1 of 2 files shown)


sogs/model.py
```diff
@@ -473,3 +511,35 @@ def get_message_deprecated(room_id, since, limit=256):
                 }
             )
     return msgs
+
+
+def ban_user(mod_user, room, ban_user):
+    """as a moderator mod_user ban a user from a room
+    return True if the user was banned, otherwise return False
+    """
```
jagerman: This would be better as a method on `User`, so that you do:

`user.ban(room, bad_user)`

sogs/legacy_routes.py
```diff
@@ -416,50 +416,23 @@ def ban_checks():
...
 
 @app.post("/legacy/ban_and_delete_all")
 def handle_legacy_banhammer():
     user, room, to_ban = ban_checks()
+    if not model.ban_user(user, room, to_ban):
+        abort(http.FORBIDDEN)
 
     with db.conn as conn:
-        apply_ban(conn, user, room, to_ban)
-
```
jagerman: This is changing the behaviour here: we want to do the ban + delete in a single transaction, so that either it all happens, or none of it happens.  Now we can get a ban applied but the posts not deleted for some reason, since they aren't happening in the same transaction.

sogs/model.py
majestrate: we need to be able to ban users from the bot api when we do not have a corresponding user in the database

sogs/legacy_routes.py
majestrate: would making this ban_user function optionally delete all posts too work?

sogs/events.py
```diff
@@ -0,0 +1,155 @@
...
+    """ find a user """
+    if metric == 'id':
+        user = model.User(id=int(query))
+    elif metric == 'pubkey':
+        user = model.User(session_id=query)
+    else:
+        raise Exception("invalid query metric: {}".format(metric))
+    reply = dict()
+    for attr in model.User.ALL_PROPS:
+        reply[attr] = getattr(user, attr)
```
jagerman: Composition:
```suggestion
    reply = {a: getattr(user, a) for a in model.User.ALL_PROPS}
```

sogs/events.py
```diff
@@ -0,0 +1,155 @@
...
+        room = model.Room(id=int(query))
+    elif metric == 'token':
+        room = model.Room(token=query)
+    else:
+        raise Exception("invalid query metric: {}".format(metric))
+    reply = dict()
+    reply['mods'] = room.get_mods()
+    reply['active_users'] = room.active_users()
+    for attr in model.Room.ALL_PROPS:
+        reply[attr] = getattr(room, attr)
```
jagerman: Composition:
```suggestion
    reply = {a: getattr(room, a) for a in model.Room.ALL_PROPS}
    reply['mods'] = room.get_mods()
    reply['active_users'] = room.active_users()
```

sogs/model.py
```diff
@@ -288,6 +314,16 @@ def __init__(self, row=None, *, id=None, session_id=None, autovivify=True, touch
             if touch:
                 self._touch()
 
+    @property
+    def rooms(self):
+        result = list()
+        """ return the rooms this user is in by their id """
```
jagerman: Docstring needs to be first

sogs/model.py
```diff
@@ -288,6 +314,16 @@ def __init__(self, row=None, *, id=None, session_id=None, autovivify=True, touch
...
+        with db.conn as conn:
+            rows = conn.execute("SELECT room FROM room_users WHERE user = ?", (self.id,))
+            for row in rows:
+                result.append(row[0])
```
jagerman: No need to wrap a single SELECT in a transaction (aside from transaction overhead it also makes it difficult to call this if you are already in a transaction).

```suggestion
        rows = db.conn.execute("SELECT room FROM room_users WHERE user = ?", (self.id,))
        result = [row[0] for row in rows]
```

But I wonder if this wouldn't be more versatile if it returned Rooms objects instead of ids:
```suggestion
        rows = db.conn.execute("""
            SELECT rooms.* FROM room_users JOIN rooms ON room_users.room = rooms.id
            WHERE user = ?
            """,
            (self.id,)
        )
        result = [Room(row=row) for row in rows]
```

sogs/model.py
jagerman: It might also be better to actual ban query into `Room` so that you can do:

`room.ban(bad_user)`

and

`user.ban(room, bad_user)`

would do the checks here then invoke `room.ban(bad_user)` if things are allowed.  (That, in particular, makes it easy to add bans from the CLI).

majestrate: i feel like inverting the user.ban calling would be better. 

```python
user.ban(mod=mod_user) # ban user from the entire server as mod
user.ban(room=room, mod=mod_user) # ban user from room as mod
user.ban(room=room) # ban user from room unconditonally 
user.ban() # ban user from entire server unconditionally
```

sogs/legacy_routes.py
jagerman: Feels kind of wrong to do two things at once in one method like that.

sogs/model.py
majestrate: another thought, we could do this: 

```python
with user.mod() as mod_user:
    mod_user.ban(user=other_guy, room=room)
```

sogs/legacy_routes.py
jagerman: Actually the transaction here is probably okay, in this order: if you ban but the delete fails you can try again.  (In the opposite order -- i.e. delete worked but ban didn't) it's difficult to actually repeat the ban.

sogs/base_config.py
```diff
@@ -70,3 +70,6 @@
 # If true, show recent messages for public rooms when accessed via a web browser.  If false only
 # show the QR code and URL but no recent messages.
 HTTP_SHOW_RECENT = True
+
+# sogs bot api listener addresses
+API_ADDRS = ['ipc://bot-api.sock']
```
jagerman: Obsolete?

sogs/filtration.py
```diff
@@ -5,9 +5,15 @@
 
 def should_drop_message_with_body(body):
     """return true if we should drop a message given its body"""
+    body = message_body(body)
```
jagerman: This should come *inside* the if for the file existing.

sogs/events.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+    _bot_category.add_request_command(
+        'find', lambda msg: _handle_request(_handle_find_request, msg)
+    )
+    _bot_category.add_request_command('ban', lambda msg: _handle_request(_handle_mod_ban, msg))
+
+
+# set up event notifiers
+for ev in EVENTS:
+    setattr(sys.modules[__name__], ev, lambda *args: _propagate_event(ev, *args))
```
jagerman: Rather than setting globals, use something like:
```Python
class Events:
    pass

event = Events()

for ev in EVENTS:
    setattr(evt, ev, lambda *args: _propagate_event(ev, *args))
```

And the users can `from .events import event` and use `event.message` or whatever.

sogs/base_config.py
majestrate: yes

sogs/events.py
```diff
@@ -0,0 +1,158 @@
...
+        raise Exception("user not banned")
+
+
+def _propagate_event(eventname, *args):
+    """ propagate an event to everyone who cares about it """
+    assert event_name_valid(eventname)
+    global _mq, _pools
+    assert _mq is not None
+    for conn in _pools[eventname]:
+        _mq.command('sogs.event', conn, eventname, *args)
```
jagerman: This would be a bit more convenient if we did `*(bt_serialize(a) for a in args)` so that the caller can pass in a dict/list/whatever.

sogs/filtration.py
majestrate: i check outside because if the body is invalid format we should drop it regardless

sogs/events.py
```diff
@@ -0,0 +1,158 @@
...
+status_OK = 'OK'
+status_ERR = 'ERROR'
+
+# the events we are able to subscribe to
+EVENTS = ('message', 'joined', 'parted', 'banned', 'unbanned', 'deleted', 'uploaded')
+
+
+def event_name_valid(eventname):
+    """ return True if this event name is something well formed """
+    return eventname.lower() in EVENTS
```
jagerman: Is the `.lower()` actually desirable here?

sogs/events.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def _user_from_conn(conn, automod=True):
+    """ make a model.User from a connection using it's curve pubkey as the session id, if automod is True we auto promote the bot to a moderator if it isn't already"""
```
jagerman:
```suggestion
    """
    make a model.User from a connection using it's curve pubkey as the session id, if automod is
    True we auto promote the bot to a moderator if it isn't already
    """
```

sogs/events.py
```diff
@@ -0,0 +1,158 @@
...
+    user = User(session_id='05{}'.format(conn.pubkey.encode('hex')))
```
jagerman:
```suggestion
    user = User(session_id='05' + conn.pubkey.encode('hex'))
```

sogs/events.py
```diff
@@ -0,0 +1,160 @@
...
+        raise Exception("user not banned")
+
+
+def _propagate_event(eventname, *args):
+    """ propagate an event to everyone who cares about it """
+    assert event_name_valid(eventname)
+    global omq, _pools
+    ev = (eventname,) + args
+    for conn in _pools[eventname]:
+        omq.send(conn, 'sogs.event', *ev)
```
jagerman: I think these should be `'sogs.' + event` rather than passing the event as the first argument, so that the caller just sets up handlers for each event rather than having to also add its own event dispatching function.

sogs/events.py
```diff
@@ -0,0 +1,160 @@
...
+    parts = msg.data()
+    if len(parts) == 0:
+        raise Exception("no events selected")
+
+    for name in parts:
+        name = name.decode()
+        if name in ('all', '*'):
+            # subscribe to all events
+            for ev in EVENTS:
+                _sub_conn_to_event(ev, msg.conn)
```
jagerman: I don't think this should be allowed because, if it's used, a future SOGS update could add new events that you don't know how to handle.

sogs/filtration.py
```diff
@@ -5,9 +5,15 @@
...
+    if body is None:
+        return False
```
jagerman: Needs a "Body failed to parse" warning here in case something changes that breaks message parsing in the future.

sogs/events.py
majestrate: probably not

sogs/events.py
majestrate: what about `event.*` ?

jagerman: Yeah, that sounds better.

sogs/filtration.py
majestrate: protobuf already does this and it is annoying af

sogs/events.py
majestrate: i agree

sogs/events.py
```diff
@@ -0,0 +1,160 @@
...
+def event_name_valid(eventname):
+    """ return True if this event name is something well formed """
+    return eventname in EVENTS
+
+
+def _user_from_conn(conn):
+    """
+    make a model.User from a connection using it's curve pubkey as the session id.
+    """
+    return model.User(session_id='05' + hexlify(conn.pubkey).decode())
```
jagerman: `bytes` have a `hex()` method:
```suggestion
    return model.User(session_id='05' + conn.pubkey.hex())
```

sogs/events.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def event_name_valid(eventname):
+    """ return True if this event name is something well formed """
+    return eventname in EVENTS
+
+
+def _user_from_conn(conn):
+    """
+    make a model.User from a connection using it's curve pubkey as the session id.
```
jagerman:
```suggestion
    make a model.User from a connection using its curve pubkey as the session id.
```

dillfrescott:
I will be able to send messages via curl with this right?

majestrate:
> I will be able to send messages via curl with this right?

no, you can do that already using the non legacy http endpoints. this PR is for using zmq.

majestrate:
the current planned http endpoints are defined here: https://github.com/oxen-io/session-pysogs/blob/dev/api.yaml
some are implemented, most are not. if you want to help implement some of the endpoints that'd be useful.

jagerman:
> some are implemented, most are not.

Actually almost all are implemented now with the recent PRs; I think the only one left is file downloading, which I'll be tackling tomorrow.

dillfrescott:
Awesome! I'm really naive when it comes to api stuff but I find it very interesting.

dillfrescott:
I just want to be able to send a message via curl lol

jagerman:
Curl likely won't ever be possible (by itself) because you have to be able to compute blake2b hashes to sign your request header with your session ID.  So a few lines of Python, sure, but it needs a bit more than a static HTTP request.

dillfrescott:
Oh. Could you take a look at this service https://github.com/caronc/apprise/wiki#notification-services and let me know if it would be possible to use it to send a message to my open group?

majestrate:
> Oh. Could you take a look at this service https://github.com/caronc/apprise/wiki#notification-services and let me know if it would be possible to use it to send a message to my open group?

make an issue for this so it is tracked. thanks.

majestrate:
> > dont use assert, throw
> 
> Why?

it is an exceptional case and should not just be an assert so that i can happen on release mode too.

PR #11, opened 2021-11-26
Session stale token workaround

jagerman:
Session requires fields be present even for an error response, which
feels wrong, but is required to get current Session to properly request
a new token.


majestrate (approved):
this is retarded.

PR #15, opened 2021-11-28
apply room post id hax for message deletion

majestrate:
* make application of room post id workarround a utils function
* apply room post id workaround for deleted message ids too

i added this to attempt to squash the deletion bug, will test on my sogs on monday.

### DIFF
--- a/sogs/model.py
+++ b/sogs/model.py
@@ -534,6 +534,7 @@ def add_post_to_room(user_id, room_id, data, sig, rate_limit_size=5, rate_limit_
 
 
 def get_deletions_deprecated(room_id, since):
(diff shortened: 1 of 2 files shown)


sogs/model.py
```diff
@@ -534,6 +534,7 @@ def add_post_to_room(user_id, room_id, data, sig, rate_limit_size=5, rate_limit_
 
 
 def get_deletions_deprecated(room_id, since):
+    since = utils.maybe_apply_post_id_hax(room_id, since)
```
jagerman, line 537: This isn't the right approach: in old SOGS, deletion ids are *completely* independent of message IDs; but also the current problem with ban-and-delete-all is with post-import messages where this adjustment isn't going to do anything.

PR #17, opened 2021-11-30
Add oxenmq as a uwsgi mule

jagerman:
This adds an oxenmq as a mule that listens on :22028 by default (using
curve encryption with the same x25519 key as the server).  This oxenmq
now handles db cleanup as well (instead of scheduling it through a
regular front-end worker).

You can kick the mule by sending it signals through the
`signal.send_signal` function so that it can be triggered when things
happen from model; a couple stubs are included here for message creation
and message deletion.

(NB: requires pyoxenmq built against the last oxenmq (1.2.9), such as
the updated python3-oxenmq 1.0.0-3 package, or installed with pip3 with
an updated liboxenmq-dev installed).


majestrate:
is this ready?

jagerman:
> is this ready?

Yes.

sogs/omq.py
```diff
@@ -0,0 +1,63 @@
...
+
+from . import crypto, config
+from .postfork import postfork
+
+omq = None
+mule_conn = None
+
+
+def make_omq():
+    omq = oxenmq.OxenMQ(privkey=crypto._privkey.encode(), pubkey=crypto.server_pubkey.encode())
```
majestrate, line 15: the public keys will be the server's pubkey when they come from the http handlers, is that desired?

majestrate (approved):

sogs/omq.py
jagerman: For workers it doesn't matter (they aren't using a curve connection anyway); for the mule, yes, this is desired.

PR #18, opened 2021-12-03
Redo how db.conn and transactions work

jagerman:
- `db.conn` is now gone; instead we have `db.get_conn()` which returns a thread-local database connection that connects on first use, though this `db.get_conn()` typically isn't used directly in favour of the following.  The thread-local is needed because otherwise (on the bot api PR branch) the mule's cleanup timer would fail because sqlite connections aren't supposed to be used across threads.

- `db.cur()` returns a cursor on the current thread's database connection (connecting first, if necessary).  This doesn't start a transaction, and so is equivalent to places where we were directly executing on db.conn outside a transaction before.

- `db.execute()` is a wrapper around getting a cursor, executing on it, and then returning it (and is basically the drop-in replacement for `db.conn.execute(...)`).  As with db.cur(), this doesn't start a transaction.

- `with db.tx() as cur:` is roughly the same as the old code:

      with db.conn as conn:
          cur = conn.cursor()

  but with two differences: the thread-local connection (as above), and more significantly, it now implements savepoints on entry and savepoint release on exit (rather than using the built-in un-nestable sqlite context manager) so that we can safely nest transactions now.

### DIFF
--- a/sogs/db.py
+++ b/sogs/db.py
@@ -1,15 +1,84 @@
 from . import config
-from .postfork import postfork
 import os
 import sqlite3
 import logging
+import threading
 
-
-conn = None
 HAVE_FILE_ID_HACKS = False
 # roomid => (max, offset).  Max is the highest message id that was in the old table; offset is the
 # value we add to ids <= that max to calculate the new database message id.
 ROOM_IMPORT_HACKS = {}
+
+
+_conns = threading.local()
+
+
+def get_conn():
+    """
+    Returns a thread-local database connection, establishing the first time it is called within a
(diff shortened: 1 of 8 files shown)


sogs/db.py
```diff
@@ -3,15 +3,86 @@
...
 
+_conns = threading.local()
+
+
+def get_conn():
+    """
+    Returns a thread-local database connection, establishing the first time it is called within a
+    thread.  This typically does not need to be called, instead use `cur()` or `tx()`.
+    """
+    if not hasattr(_conns, 'db'):
```
majestrate: `'db'` or `'conn'` ?

majestrate (approved):

PR #19, opened 2021-12-06
Implement whisper in backend + retrieval

jagerman:


sogs/db.py
```diff
@@ -144,7 +144,46 @@ def database_init():
             )
             raise
 
-    # Any future migrations go here
+    # New columns that might need to be added:
```
majestrate, line 159: i would put this new hunk in its own function to mark it as a distinct new discrete unit of migration.

sogs/routes.py
```diff
@@ -88,33 +88,11 @@ def get_recent_room_messages(room):
...
-
-    return jsonify(msgs)
+    # FIXME: this is temporary, for the basic front-end; for proper implementation we should have a
+    # user by this point.
+    user = None
+
+    if not room.check_permission(None, read=True):
+        abort(http.FORBIDDEN)
+
+    return utils.jsonify_with_base64(room.get_messages_for(None, recent=True, limit=limit))
```
majestrate: change `None` to `user` in the 2 places it is used here?

sogs/utils.py
```diff
@@ -53,6 +54,32 @@ def decode_hex_or_b64(data: bytes, size: int):
...
+    """
+    return json.dumps(_json_b64_impl(val))
+
+
+def jsonify_with_base64(val):
+    """
+    Returns a flask response set up for json (like flask.jsonify(...)), but uses json_with_base64
+    for the encoding.
+    """
+    return Response(json_with_base64(val), mimetype="application/json")
```
majestrate, line 80: why not just `return flask.jsonify(json_with_base64(val))` ?

majestrate: rather... `jsonify(_json_b64_impl(val))`

PR #20, opened 2021-12-08
Model consolidation

jagerman:
(Build on top of #19)

### DIFF
--- a/sogs/db.py
+++ b/sogs/db.py
@@ -1,4 +1,5 @@
 from . import config
+from . import crypto
 import os
 import sqlite3
 import logging
@@ -131,7 +132,11 @@ def database_init():
     migrate_v01x(conn)
     add_new_columns(conn)
     update_whisper_views(conn)
+    create_message_details_deleter(conn)
     check_for_hacks(conn)
+
+    # Make sure the system admin users exists
+    create_admin_user(conn)
 
     conn.close()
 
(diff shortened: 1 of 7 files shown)


sogs/model.py
```diff
@@ -541,14 +1065,33 @@ def set_moderator(self, *, admin=False, visible=False):
         self.global_moderator = True
         self.visible_mod = visible
 
-    def remove_moderator(self):
+    def remove_moderator(self, *, removed_by: User):
         """Removes this user's global moderator/admin status, if set."""
+
+        if not removed_by.global_admin:
```
majestrate, line 1071: what if removed_by is None?

jagerman: Then you called it wrong and deserve the exception that you get.

sogs/db.py
```diff
@@ -198,6 +204,26 @@ def update_whisper_views(conn):
...
+    if not conn.execute(
+        """
+        SELECT EXISTS(
+            SELECT 1 FROM sqlite_master WHERE type = 'trigger' AND name = 'message_details_deleter'
+        )
+        """
+    ).fetchone()[0]:
+        conn.execute(
+            """
+CREATE TRIGGER message_details_deleter INSTEAD OF DELETE ON message_details
```
majestrate: `CREATE TRIGGER IF NOT EXISTS message_details_deleter ...`

majestrate (approved):

PR #21, opened 2021-12-09
Move config into sogs.ini

jagerman:
Moves all the config settings that are usefully settable into sogs.ini.  (Some of the ones which aren't really config settings -- like the schema path and local filename restrictions -- are just kept as constants without being ini configurable).

By default this loads config from sogs.ini, but can load from an arbitrary path if the SOGS_CONFIG environment variable is set (for example, using `env = SOGS_CONFIG=/some/path` in the uwsgi config).


sogs.ini.sample
```diff
@@ -0,0 +1,102 @@
...
+; communication between sogs processes.  This is unauthenticated and so must not be a publicly
+; accessible address!
+;
+;omq_internal = ipc://./omq.sock
+
+
+; Whether we should show recent messages for public rooms on the sogs room page when visited by a
+; browser.
+;
+;http_show_recent = yes
```
majestrate, line 48: maybe also an option like `enable_room_viewer` for disabling it entirely as well?

jagerman: Sure, but out of scope for this PR, which is just about moving the settings we currently have.

sogs/config.py
```diff
@@ -1,5 +1,126 @@
...
+    # enabled).
+    if 'log' in cp.sections() and 'level' in cp['log']:
+        app.logger.setLevel(cp['log']['level'])
+
+    # Map of: section => { param => ('GLOBAL', test lambda, value lambda) }
+    # global is the string name of the global variable to set
+    # test lambda returns True/False for validation (if None/omitted, accept anything)
+    # value lambda extracts the value (if None/omitted use str value as-is)
+    setting_map = {
+        'db': {'url': ('DB_PATH', lambda x: x.startswith('sqlite:///'), lambda x: x[10:])},
```
majestrate, line 62: should it be `sqlite://` instead here?

jagerman: No, the triple slashes is correct (for future SQL alchemy url support).  Formally it is `dialect://host/stuff` but sqlite has no host.

jagerman: (and for an absolute path it's `sqlite:////full/path`)

majestrate: ew, ok.

majestrate (approved):

PR #26, opened 2021-12-16
Profanity

jagerman:
better_profanity has a fairly extensive built-in bad word list and handles lots of things like word permutations.

This also adds the ability to silently filter: if enabled (which it is, by default) then profanity is inserted but not distributed, so the poster gets a successful post (which requires we return an actual message id), but the message goes nowhere.  (It is still in the database, though, which means we have the ability to let admins see filtered messages).

This commit also fixes bugs in our message handling:
- open group messages a protobuf Content, not DataMessage, and so we were parsing it wrong (and getting invalid utf-8 data errors).
- we were not stripping padding correctly, nor re-adding padding correctly (but the first bug masked the second one) and so were still storing full-padding messages.


sogs.ini.sample
```diff
@@ -1,6 +1,7 @@
 ; SOGS configuration.  Copy this file to sogs.ini and edit as desired.  It can be left mostly (or
 ; entirely) empty: defaults are shown in the comments.
 
+
```
majestrate, line 4: stray white space addition that will haunt our souls for several years? :thinking:

majestrate (approved):

sogs.ini.sample
jagerman: To be consistent; all the other ini sections have 2 spaces.

PR #31, opened 2021-12-21
Hide system user from get_all_global_moderators

jagerman:
The system user appears in `python3 -msogs -M`, which might worry an
admin.  This hides it (since it isn't really an admin, but rather just
a stub for internal tasks).

(Thanks to @ianmacd for the report).


jagerman:
(This also showed up in the room mod list returned to admins; found and fixed in #34)

PR #33, opened 2022-01-06
PostgreSQL support

jagerman:
Working postgresql support.  Builds on #32


jagerman:
Some comments:

- Adds a schema.pgsql that does the same as the sqlite version, but with postgresql specifics.
- `user` is a reserved SQL keyword, so had to get quoted as `"user"` in the queries.
- fixes a couple other queries that postgresql didn't like.
- postgresql doesn't support `.lastrowid`, but does support RETURNING, so this adds an abstract function for inserting and fetching the primary key using RETURNING into db.py.  (For sqlite, we only use it for sqlite 3.35+, and otherwise fall back to using `.lastrowid`).
- pg `bytea`s are returned as memoryviews, which broke some of the utils code expecting `bytes`.

PR #38, opened 2022-01-18
 New authentication

jagerman:
This adds replay-protected, one-shot request authentication, intended to replace the legacy request token interface.

Includes tests.

See the comments at the top of sogs/routes/auth.py for details on how it all works.

(Builds on top of #36 )


sogs/routes/auth.py
```diff
@@ -0,0 +1,236 @@
...
+# authentication mechanism; in order to call them you must invoke them with a leading `/legacy/`
+# prefix (e.g. `GET /legacy/rooms`).
+
+
+def abort_with_reason(code, msg):
+    app.logger.warning(msg)
+    abort(Response(msg, status=code, mimetype='text/plain'))
+
+
+def require_user():
```
majestrate, line 107: decorator?

sogs/routes/auth.py
```diff
@@ -0,0 +1,236 @@
...
+    try:
+        query('INSERT INTO user_request_nonces ("user", nonce) VALUES (:u, :n)', u=user.id, n=nonce)
+    except sqlalchemy.exc.IntegrityError:
+        abort_with_reason(http.TOO_EARLY, "Invalid authentication: X-SOGS-Nonce cannot be reused")
+
+    # Signature validation
+
+    # shared_key is hash of a*B || A || B = b*A || A || B where b/B is the server keypair and A is
+    # the session id pubkey.
+    shared_key = nacl.hash.blake2b(
```
majestrate: i feel like all of the cryptography details should be in crypto.py

sogs/routes/auth.py
```diff
@@ -0,0 +1,236 @@
...
+        crypto_scalarmult(crypto._privkey.encode(), A) + A + crypto.server_pubkey_bytes,
+        digest_size=64,
+        salt=nonce,
+        person=b'sogs.shared_key',
+        encoder=RawEncoder,
+    )
+
+    hasher = nacl.hashlib.blake2b(
+        request.method.encode() + request.path.encode(),
+        digest_size=64,
```
majestrate: make this 32 bytes instead so the final digest is a bit more palatable

sogs/routes/auth.py
jagerman: Added sogs.hashing for the blake2b calls; the crypto details in here, though, seem non-generic enough and with details like working around the flask full_path bug that it makes more sense to leave them here.

sogs/hashing.py
```diff
@@ -0,0 +1,42 @@
+import nacl.hashlib
+from typing import Union, Iterable
+
+
+def blake2b(
+    data: Union[bytes, Iterable[bytes]],
```
majestrate: what if this was `Iterable` of things implementing the buffer interface or whatever the base type is called that is accepted by hasher.update?

sogs/routes/auth.py
jagerman: Made it 42 instead (so that it's distinct from common 32-byte values like pubkeys, b64-encodes without padding, and isn't so long).

sogs/hashing.py
jagerman: That will work fine as is, so just removed the type hint.

majestrate (approved):

PR #45, opened 2022-01-27
Add support for IN clauses to db.query()

jagerman:
Expanding IN clauses in bindings with SQLAlchemy requires a little more
work (and the existing IN clauses were broken).

This expose the needed mechanics in the query() interface, used such as:
```Python
    db.query(
        "SELECT * FROM blah WHERE id IN :ids",
        ids=[1,2,3],
        bind_expanding=['ids']
    )
```

### DIFF
--- a/sogs/db.py
+++ b/sogs/db.py
@@ -5,6 +5,7 @@ import os
 import logging
 import importlib.resources
 import sqlalchemy
+from sqlalchemy.sql.expression import bindparam
 
 HAVE_FILE_ID_HACKS = False
 # roomid => (max, offset).  Max is the highest message id that was in the old table; offset is the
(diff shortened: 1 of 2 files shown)


jagerman:
The IN clauses fixed here aren't currently being hit by the test suite, but I'll address that in one of the other incoming endpoint PRs.

majestrate:
do we have tests for this yet?

jagerman:
> do we have tests for this yet?

Incoming PRs are using this.  Tests for the currently broken ones likely fixed here will be coming shortly as well (when those endpoints get implemented/testsed).

I wanted to pull this into its own PR, though, as it isn't really related to any added endpoints.

majestrate (approved):
yeah sure look good

PR #46, opened 2022-01-28
Moderator/admin control endpoints

jagerman:
Adds endpoints for adding/removing moderators/admins from rooms/server.

### DIFF
--- a/sogs/routes/users.py
+++ b/sogs/routes/users.py
@@ -1,0 +1,115 @@
+from .. import db, http
+from ..model import room as mroom
+from ..model.user import User
+from ..web import app
+from . import auth
+
+from flask import abort, jsonify, g, Blueprint, request
+
+# User-related routes
+
+
+users = Blueprint('users', __name__)
+
+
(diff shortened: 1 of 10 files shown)


sogs/routes/users.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+
+    # Filter out any invalid or redundant arguments:
+    if (admin, mod) == (None, None):
+        app.logger.warning(
+            "Invalid moderator request: at least one of admin/moderator must be specified"
+        )
+        abort(http.BAD_REQUEST)
+    elif (admin, mod) == (True, False):
+        app.logger.warning("Invalid moderator call: admin=True, moderator=False is impossible")
```
majestrate, line 40: this should be error level

jagerman: Leaving as warning because this is a client-side error, but still wanted to log it in case a server admin is trying to figure out why some (wrong) client call "doesn't work".

majestrate (approved):

PR #47, opened 2022-01-28
Bans, unbans

jagerman:
Allows adding bans, ban timeouts, and unbanning at both the room and server level.

Timeouts revealed a deficiency in the user permission futures table that would be difficult to deal with, so I split up permission futures and ban futures into two separate tables which makes things quite a bit simpler -- *and* allows global bans to have an expiry, which wasn't possible before.

### DIFF
--- a/sogs/routes/users.py
+++ b/sogs/routes/users.py
@@ -12,6 +12,64 @@ from flask import abort, jsonify, g, Blueprint, request
 users = Blueprint('users', __name__)
 
 
+def extract_rooms_or_global(req, admin=True):
+    """
+    Extracts the rooms / global parameters from the request body checking them for validity and
+    expanding them as appropriate.
+
+    Throws a flask abort on failure, returns (rooms, global) which will be either ([list of Rooms],
+    None) for a room operation or (None, True) for a global operation.
+
+    admin specifies whether we require admin permission (if True) or just moderator permission,
+    either in all rooms specified, or globally.  (Similarly it affects what `rooms=['*']` expands
+    to).
+    """
+
+    room_tokens, global_ = req.get('rooms'), req.get('global', False)
+
+    if room_tokens and not isinstance(room_tokens, list):
+        app.logger.warning("Invalid request: rooms must be a list")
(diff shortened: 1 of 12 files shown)


sogs/routes/users.py
```diff
@@ -98,18 +125,59 @@ def set_mod(sid):
...
                 user.set_moderator(added_by=g.user, admin=bool(admin), visible=visible)
 
     return jsonify({})
+
+
+@users.post("/user/<SessionID:sid>/ban")
+@auth.user_required
+def ban_user(sid):
+
+    user = User(session_id=sid)
```
majestrate, line 150: `autovivify=False` ?

sogs/routes/users.py
```diff
@@ -98,18 +125,59 @@ def set_mod(sid):
...
+        user.ban(banned_by=g.user, timeout=timeout)
+
+    return {}
+
+
+@users.post("/user/<SessionID:sid>/unban")
+@auth.user_required
+def unban_user(sid):
+
+    user = User(session_id=sid)
```
majestrate, line 173: `autovivify=False` here too?

tests/test_user_routes.py
```diff
@@ -470,3 +468,230 @@ def room_mods(u=user2):
     sogs_post(client, url_u1_mod, {'rooms': ['test-room'], 'admin': True}, global_admin)
     assert room_mods() == ([s_mod], sorted([s_admin, s_user]), [], [])
     assert room_mods(mod) == ([s_mod], sorted([s_admin, s_user]), [], [s_gadmin])
+
+
+def test_bans(client, room, room2, user, user2, mod, global_mod):
```
majestrate, line 473: case for banning non existent user?

sogs/routes/users.py
jagerman: It needs to get inserted if it doesn't exist, so that I can add the ban row (so that you can preemptively ban someone who has never actually hit the server yet).

majestrate: ok

majestrate (approved):

PR #48, opened 2022-01-29
Room info update endpoint

jagerman:
Adds `PUT /room/{token}` endpoint to accept a new name, description, and default_{read,write,upload} settings.

Also fixes some incorrect returned values (including fields with `false` values instead of omitting them), and exposes the default_read/write/upload values to room info for moderators.

### DIFF
--- a/sogs/routes/auth.py
+++ b/sogs/routes/auth.py
@@ -123,6 +123,44 @@ def user_required(f):
     return required_user_wrapper
 
 
+def require_mod(room, *, admin=False):
+    """Checks a room for moderator or admin permission; aborts with 401 Unauthorized if there is no
+    user in the request, and 403 Forbidden if g.user does not have moderator (or admin, if
+    specified) permission."""
+    require_user()
+    if not (room.check_admin(g.user) if admin else room.check_moderator(g.user)):
+        abort_with_reason(
+            http.FORBIDDEN,
+            f"This endpoint requires {'admin' if admin else 'moderator'} room permissions",
+        )
+
+
+def mod_required(f):
+    """Decorator for an endpoint that requires a user that has moderator permission in the given
+    room.  The function must take a `room` argument by name, as is typically used with flask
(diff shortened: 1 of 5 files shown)


sogs/routes/rooms.py
```diff
@@ -58,6 +66,65 @@ def get_rooms():
     return jsonify([get_one_room(r) for r in mroom.get_readable_rooms(g.user)])
 
 
+BAD_NAME_CHARS = {c: None for c in range(32)}
+BAD_DESCRIPTION_CHARS = {c: None for c in range(32) if not (0x09 <= c <= 0x0A)}
+
+
+@rooms.put("/room/<Room:room>")
+@auth.user_required
```
majestrate: feel like we could have a `@auth.require_admin` for this

jagerman: ~~I'm not sure how to make that work through a decorator, because it isn't just global admins but also room admins (which means we need to be able to call `room.check_admin(g.user)`, and so need the `room` argument).~~

jagerman: Added.

sogs/routes/auth.py
```diff
@@ -123,6 +123,41 @@ def required_user_wrapper(*args, **kwargs):
     return required_user_wrapper
 
 
+def require_mod(room, *, admin=False):
+    """Checks a room for moderator or admin permission; aborts with 401 Unauthorized if there is no
+    user in the request, and 403 Forbidden if g.user does not have moderator (or admin, if
+    specified) permission."""
+    require_user()
+    if not (room.check_admin(g.user) if admin else room.check_moderator(g.user)):
+        abort_with_reason(http.FORBIDDEN, 'This endpoint requires moderator permissions')
```
majestrate: error string should contain 'admin' not 'moderator' when it's looking for admin not moderator

jagerman: Fixed.

majestrate (approved):
look fine

PR #49, opened 2022-01-31
Add new config Boolean, http_show_index, for room directory display.

ianmacd:
'http_show_index = No' disables the default behaviour of displaying a
directory of public rooms when the root server URL is visited using a
browser.

### DIFF
--- a/sogs/routes/__init__.py
+++ b/sogs/routes/__init__.py
@@ -30,6 +30,8 @@ def serve_index():
     rooms = get_readable_rooms()
     if len(rooms) == 0:
         return render_template('setup.html')
+    if not config.HTTP_SHOW_INDEX:
+        abort(http.FORBIDDEN)
     return render_template(
         "index.html", url_base=config.URL_BASE, rooms=rooms, pubkey=crypto.server_pubkey_hex
     )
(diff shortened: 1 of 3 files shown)


sogs/routes/__init__.py
```diff
@@ -30,6 +30,8 @@ def serve_index():
     rooms = get_readable_rooms()
     if len(rooms) == 0:
         return render_template('setup.html')
+    if not config.HTTP_SHOW_INDEX:
+        abort(http.FORBIDDEN)
```
majestrate, line 34: would `http.NOT_FOUND` be more appropriate?

majestrate:
FYI: this changeset does not prevent anonymous read only access, it just disables serving the room viewer page (which uses other API endpoints via ajax to get posts).

majestrate:
sorry, it only disables the room list page, room viewer still works if you have the room name. is this desirable?

sogs/routes/__init__.py
jagerman: Forbidden is what web servers return when directory indexing is turned off and you try to access a directory, so seems appropriate here.

ianmacd:
> sorry, it only disables the room list page, room viewer still works if you have the room name. is this desirable?

I think so, because the room viewer already has a config option that can be used to separately disable that: `http_show_recent`.

majestrate (approved):

PR #50, opened 2022-01-31
user inbox endpoints

majestrate:
endpoints for user to user control messages


api.yaml
```diff
@@ -755,6 +755,15 @@ paths:
...
         404:
           $ref: "#/paths/~1room~1%7BroomToken%7D~1file~1%7BfileId%7D/get/responses/404"
+  /user/inbox:
+    get:
+      tags: [Users]
+      summary: get all of the current user's inbox messages
+  /user/{sessionId}/inbox:
+    post:
+      tags: [Users]
+      summary: Submit an inbox message to another user
```
jagerman: Can use some more docs on what will be in the response.  It's definitely a bit tedious and massively over-indented but copy-and-pasting from one of the other ones works okay.

sogs/db.py
```diff
@@ -214,7 +214,35 @@ def add_new_tables(conn):
...
         added = True
-
+    if 'inbox' not in metadata.tables:
+        if engine.name == 'sqlite':
+            conn.execute(
+                """
+CREATE TABLE inbox (
+    id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    recipient BIGINT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+    sender BIGINT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
```
jagerman:
```suggestion
    id INTEGER PRIMARY KEY,
    recipient INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
    sender INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
```

sogs/model/message.py
```diff
@@ -0,0 +1,65 @@
+from .. import db
+from ..db import query
+
+from .user import User
+
+
+class Message:
+    """Class representing a DM between users
```
jagerman, line 10:
```suggestion
    """
    Class representing a DM between users
```

api.yaml
majestrate: yes, coming soon ™

sogs/model/message.py
```diff
@@ -0,0 +1,65 @@
...
+        Constructs a Message from a pre-retrieved row *or* sender recipiant data and signature.
+        """
+        if row is None:
+            if None in (sender, recip, data, signature):
+                raise ValueError(
+                    "Message() error: provided no row but did not provide all members for construction"
+                )
+            for arg in (sender, recip):
+                if not isinstance(arg, User):
+                    raise ValueError("Message() error: sender or recipient was not a User model")
```
jagerman:
```suggestion
            if not all(isinstance(arg, User) for arg in (sender, recip)):
                raise ValueError("Message() error: sender or recipient was not a User model")
```

sogs/model/message.py
```diff
@@ -0,0 +1,65 @@
...
+
+            with db.transaction():
+                row = query(
+                    "INSERT INTO inbox(sender, recipient, data, signature) VALUES(:sender, :recipient, :data, :signature) RETURNING *",
```
jagerman: I think this is going to flake8 fail.  What I've been writing for these queries is:
```
            """
            INSERT INTO inbox (sender, recipient, data, signature)
            VALUES (:sender, ....)
            """,
```
Also needs `db.insert_and_get_pk` because `RETURNING *` isn't supported on non-super-recent sqlite.

sogs/model/message.py
```diff
@@ -0,0 +1,65 @@
...
+    def data(self):
+        return self._row['data']
+
+    @property
+    def sender(self):
+        return User(id=self._row['sender'], autovivify=False)
+
+    @property
+    def recipient(self):
+        return User(id=self._row['recipient'], autovivify=False)
```
jagerman: Would be preferable to cache the `User` object here (in `self._sender` and `self._recipient` or something like that) so that if the caller writes code that does something like:
```Python
if msg.sender == whatever:
    print(msg.sender)
```
that you only fetch it from the database once.

sogs/model/user.py
```diff
@@ -230,6 +230,14 @@ def unban(self, *, unbanned_by: User):
         app.logger.debug(f"{unbanned_by} removed global ban on {self}")
         self.banned = False
 
+    def verify(self, *, message: bytes, sig: bytes):
+        """verify signature signed by this session id
+        return True if the signature is valid otherwise return False
+        """
+        return crypto.xed25519_verify(
+            sig, utils.decode_hex_or_b64(self.session_id[2:], size=32), message
```
jagerman: The `or_b64` here seems strange.  Could this just be `bytes.from_hex(self.session_id[2:])`?  That will throw a ValueError on non-hex; I assume xed25519_verify checks argument lengths and throws if not right.

sogs/routes/users.py
```diff
@@ -70,6 +72,51 @@ def extract_rooms_or_global(req, admin=True):
...
+    n = request.query.get("count", 10, type=int)
+    msgs = [msg.info for msg in g.user.inbox(offset=offset, count=n)]
+    return jsonify(msgs)
+
+
+@users.post("/user/<SessionID:sid>/inbox")
+@auth.user_required
+def send_inbox(sid):
+    """ send a message to a recipiant user via their session id """
+    if sid[0:2] == '05' and not g.user.moderator:
```
jagerman: I don't think there's a use case for sending to a non-blinded id; we can just hard-fail for anyone who tries to send to a non-blinded address.  (Session clients shouldn't ever fetch for their true ID anyway).

sogs/routes/users.py
```diff
@@ -70,6 +72,50 @@ def extract_rooms_or_global(req, admin=True):
...
+    n = request.query.get("count", 10, type=int)
+    msgs = [msg.info for msg in g.user.inbox(offset=offset, count=n)]
+    return jsonify(msgs)
+
+
+@users.post("/inbox/<SessionID:sid>")
+@auth.user_required
+def send_inbox(sid):
+    """ send a message to a recipiant user via their session id """
+    if sid[0:2] == '05':
```
jagerman: Maybe `!= '15'`?

majestrate:
most of the substance is there, api.yaml needs to be fully filled out still.

majestrate:
post /inbox/<sessionid> takes a json object with 2 members, data and signataure, both base64 encodded strings, returns a json object with keys, id and expires at, both numbers

get /inbox takes 2 potential query paramters, offset and count. 
returns a list of count objects which represent messages, message has keys id, data, signature, sender, recipient.

jagerman:
I think we should drop `id` from the return; the sender has no use for it.

For the receiver polling their inbox, we should have a `since` argument that takes an id and returns values > that value (so that clients have a way to retrieve only new inbox messages).

sogs/db.py
```diff
@@ -223,7 +223,35 @@ def add_new_tables(conn):
...
+    if 'inbox' not in metadata.tables:
+        if engine.name == 'sqlite':
+            conn.execute(
+                """
+CREATE TABLE inbox (
+    id INTEGER PRIMARY KEY,
+    recipient INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+    sender INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+    body BLOB NOT NULL,
+    signature BLOB NOT NULL,
```
jagerman: I think we should add a:
```sql
    timestamp FLOAT DEFAULT ((julianday('now') - 2440587.5)*86400.0) /* unix timestamp */)
```

sogs/db.py
```diff
@@ -223,7 +223,35 @@ def add_new_tables(conn):
...
+    expiry FLOAT DEFAULT ((julianday('now') - 2440587.5 + 1.0)*86400.0) /* now + 24h */)
```
jagerman, line 256: 24h seems way too short

jagerman: Perhaps 7 days?  (Also the postgres version below is still defaulting to 15 days).

jagerman: Actually, we should make this admin-tunable.  i.e. keep the default here at something like 7d, but then add a config option for how many days and then explicitly specify it when inserting.

sogs/model/message.py
```diff
@@ -0,0 +1,94 @@
...
+            if None in (sender, recip, data, signature):
+                raise ValueError("Message() error: no row or data provided")
+            if not all(isinstance(arg, User) for arg in (sender, recip)):
+                raise ValueError("Message() error: sender or recipient was not a User model")
+
+            with db.transaction():
+                row = insert_and_get_pk(
+                    """
+            INSERT INTO inbox (sender, recipient, data, signature)
+            VALUES (:sender, ....)
```
jagerman: Doesn't seem quite right...

sogs/model/message.py
```diff
@@ -0,0 +1,94 @@
...
+            """,
+                    "id",
+                    sender=sender.id,
+                    recipiant=recip.id,
+                    data=data,
+                    signture=signature,
```
jagerman:
```suggestion
                    sender=sender.id,
                    recipient=recip.id,
                    data=data,
                    signature=signature,
```

sogs/model/message.py
```diff
@@ -0,0 +1,94 @@
...
+                ).first()
```
jagerman: This is going to return just the id, not the row.

~~It seems worth adding a `insert_and_get_row` into db.py that does a `RETURNING *` if `have_returning` and otherwise does a `insert_and_get_pk` followed by selecting the row and returning it.~~

Edit: I added (and pushed) a insert_and_get_row that does this now (using `RETURNING *` when available, otherwise selecting it after the insert), but haven't updated this code to use it.  Usage should be:

```Python
row = insert_and_get_row("query", "inbox", "id", sender=sender.id, etc.)
```
(i.e. drop the `.first()` off the end, too).

sogs/model/message.py
```diff
@@ -0,0 +1,94 @@
...
+        if row is None:
+            if None in (sender, recip, data, signature):
+                raise ValueError("Message() error: no row or data provided")
+            if not all(isinstance(arg, User) for arg in (sender, recip)):
+                raise ValueError("Message() error: sender or recipient was not a User model")
+
+            with db.transaction():
+                row = insert_and_get_pk(
+                    """
+            INSERT INTO inbox (sender, recipient, data, signature)
```
jagerman: We should add `, expiry` here and bind its value to something like `expiry=time.time() + config.DM_EXPIRY_DAYS*86400`

sogs/model/message.py
```diff
@@ -0,0 +1,94 @@
...
+    @property
+    def info(self):
+        return {
+            "id": self.id,
+            "expires_at": self.expires_at,
+            "data": self.data,
+            "signature": self.signature,
+            "sender": self.sender.session_id,
+            "recipient": self.recipient.session_id,
+        }
```
jagerman: This mapping feels like it belongs in routes code rather than the model.

sogs/model/message.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @property
+    def info(self):
+        return {
+            "id": self.id,
+            "expires_at": self.expires_at,
+            "data": self.data,
+            "signature": self.signature,
+            "sender": self.sender.session_id,
+            "recipient": self.recipient.session_id,
```
jagerman: Is there any reason to include this field in the response to the message retriever?

majestrate: no

sogs/model/message.py
```diff
@@ -0,0 +1,94 @@
...
+    def to(user, since=None):
+        """ get all message for a user, returns a generator """
+        if since:
+            rows = query(
+                "SELECT * FROM inbox WHERE recipient = :recip AND id > :since_id",
+                recip=user.id,
+                since_id=since,
+            )
+        else:
+            rows = query("SELECT * FROM inbox WHERE recipient = :recip", recip=user.id)
```
jagerman: This (and the `since` one, above) need an `ORDER BY id`

sogs/model/message.py
```diff
@@ -0,0 +1,94 @@
...
+                    recipiant=recip.id,
+                    data=data,
+                    signture=signature,
+                ).first()
+        # sanity check
+        assert row is not None
+        self._row = row
+
+    @staticmethod
+    def to(user, since=None):
```
jagerman: I'm thinking we should have a `limit` parameter (passed in by get parameter that works just like the limit parameters for /room/{name}/messages/since/{id}, including the default to 100).  Otherwise you might be able to fill a user's inbox with large messages that they can't retrieve at all (because the response would exceed the size limit).

sogs/model/user.py
```diff
@@ -230,9 +230,20 @@ def unban(self, *, unbanned_by: User):
...
+
+    @property
+    def is_blinded(self):
+        """True if the user's session id is a derived key"""
+        return self.session_id[0:2] == '15'
+
     @property
     def system_user(self):
-        """True iff this is the special SOGS system user created for internal database tasks"""
+        """True if this is the special SOGS system user created for internal database tasks"""
```
jagerman: That was not a typo; "iff" = shorthand for "if and only if"

sogs/model/message.py
majestrate: i want to be able to do a `[msg.info for msg in fetch_messages_however_we_do_that()]`

sogs/model/message.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        Constructs a Message from a pre-retrieved row *or* sender recipient data and signature.
+        """
+        if row is None:
+            if None in (sender, recip, data, signature):
+                raise ValueError("Message() error: no row or data provided")
+            if not all(isinstance(arg, User) for arg in (sender, recip)):
+                raise ValueError("Message() error: sender or recipient was not a User model")
+
+            with db.transaction():
```
jagerman: Not needed; `insert_and_get_row` will start one itself if it needs to do multiple queries (and if it's a single query we don't need a tx).

tests/test_dm.py
```diff
@@ -0,0 +1,34 @@
+from request import sogs_get, sogs_post
+from sogs import crypto
+
+from base64 import b64encode
```
jagerman:
```suggestion
from sogs.utils import encode_base64
```

tests/test_dm.py
```diff
@@ -0,0 +1,34 @@
...
+    r = sogs_get(client, '/inbox', banned_user)
+    assert r.status_code == 403
+
+
+def make_post(data, user, *, blind=False):
+    privkey = user.privkey.encode()
+    if blind:
+        privkey = crypto.compute_derived_key_bytes(privkey)
+    sig = crypto.xed25519_sign(privkey, data)
+    return {'message': b64encode(data).decode('ascii'), 'signature': b64encode(sig).decode('ascii')}
```
jagerman:
```suggestion
    return {'message': encode_base64(data), 'signature': encode_base64(sig)}
```

tests/test_dm.py
jagerman: The one from utils includes the trailing `.decode()` which is virtually always what you want.  `base64` just likes being obtuse in every way.

jagerman (approved):

api.yaml
```diff
@@ -1687,6 +1755,42 @@ components:
...
+          type: string
+          format: byte
+          description: >
+            The direct message data, encoded in base64.
+        signature:
+          type: string
+          format: byte
+          description: >
+            An XEd25519 signature of the data contained in `data`, signed using the X25519 pubkey
+            contained in the user's Session ID.
```
jagerman, line 1765: Thinking about this some more I think the signature is useless here: Session is going to be depositing some authenticated-encrypted data and forcing an extra signature on top of that doesn't do anything useful (especially because they've already signed the incoming request as part of the X-SOGS-* auth headers).

api.yaml
```diff
@@ -1687,6 +1755,42 @@ components:
...
+        sender:
+          allOf:
+            - $ref: "#/components/schemas/SessionID"
+            - type: object
+              description: "The session ID of the user who sent this message."
+        recipient:
+          allOf:
+            - $ref: "#/components/schemas/SessionID"
+            - type: object
+              description: "The session ID to which this message was sent."
```
jagerman, line 1780: I think this got dropped from the returned DM data

sogs/db.py
```diff
@@ -223,7 +244,37 @@ def add_new_tables(conn):
...
+    if 'inbox' not in metadata.tables:
+        if engine.name == 'sqlite':
+            conn.execute(
+                """
+CREATE TABLE inbox (
+    id INTEGER PRIMARY KEY,
+    recipient INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+    sender INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+    body BLOB NOT NULL,
+    signature BLOB NOT NULL,
```
jagerman: drop sig

sogs/db.py
```diff
@@ -223,7 +244,37 @@ def add_new_tables(conn):
...
+            conn.execute("CREATE INDEX inbox_recipient ON inbox(recipient)")
+        else:
+            conn.execute(
+                """
+CREATE TABLE inbox (
+    id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    recipient BIGINT NOT NULL REFERENCES users ON DELETE CASCADE,
+    sender BIGINT NOT NULL REFERENCES users ON DELETE CASCADE,
+    body BYTEA NOT NULL,
+    signature BYTEA NOT NULL,
```
jagerman: drop sig

jagerman (changes requested):
Need to drop signatures out of here, then squash/rewrite the commit history into fewer commits, then I think this is good to go.

PR #52, opened 2022-02-04
v4 onion requests

jagerman:
Adds a new v4 SOGS onion request that allows efficiently passing binary data.

This solves six glaring issues with v3 onion requests:
1. it is impossible to pass binary data through an onion request because onion requests are always json, and thus v3-reachable endpoints have to accept data as base64, which means a 33% increase in data size.
2. for endpoints where we want to be able to pass binary data and *aren't* going through onion requests (e.g. from an oxenmq bot) we have to duplicate the endpoint to have a second version that takes it without json encoding overhead, or else incur the overhead.
3. v3 requests always return json, and thus have the same amplification of data (due to a requirement to have to slam binary data into json) on the return path to the client.
4. v3 requests always apply *another* round of base64 encoding after the encryption for the client, which means there is a 78% amplification of any binary data being returned to a client.
5. The base64 willy nilly everywhere makes onion requests slower: because, in either direction, we have to fully transmit the data from one hop to the next before we can begin transmitting to the next one, this means that the amplification adds latency when uploading/downloading large attachments.
6. v3 requests completely discard any returned headers which mean, for instance, that serving a file with a proper Content-Disposition header specifying the filename is impossible.  (Instead, you'd have to slam everything into the json return rather than use the standard headers).

These are solved by:
1. data is passed in two parts: a json request metadata, and a binary body.
2. no duplication request handlers are needed because we don't need to avoid inefficient extra encoding
3. endpoints accessed by v4 can return any content-type they like.  For attachments, for instance, we can return `application/octet-stream` bytes with no overhead.
4. v4 requests do not base64 encode anything; they return encrypted data as binary.
5. No data amplification means no added latency
6. v4 responses include response headers (like content-type) 

## Request encoding

The quick overview of how requests get encoded with v4 is:

v4 allows more generic requests by taking the data in two parts:

- metadata json describing the request
- the request data

The metadata is effectively the same as the v3 request body, except that it does not contain a `body` element.

The request data (which is only included in POST/PUT requests) is binary data and can be anything.  Unlike v3, this means the request needs to include a Content-Type header, which typically not specified in v3 (since it could only ever be application/json).

Finally these parts are glued together using simple bencoding, chosen to only be a list of strings that is simple enough that it can be easily encoded/decoded by hand without requiring a full bencode library (though of course a full bencode parser/producer will work fine).  The format of the bencode subset used for v4 onion requests is this:

- `N:data` is a binary string, where `N` is a decimal number.  For example `11:hello world` encodes the byte string `hello world`.
- `l` begins a list; the list is followed by 1 or 2 binary strings; then is terminated with an `e`.

so, for example, a v4 onion request to `GET` the `/room/test-room` endpoint would look like this:
```
l45:{"method":"GET","endpoint":"/room/test-room"}e
```
and a POST request for `/room/test-room/pin/123` with a json body of `{}` would look like:
```
l100:{"method":"POST","endpoint":"/room/test-room/pin/123","headers":{"Content-Type":"application/json"}}2:{}e
```

## Response encoding

Response encoding is nearly identical to the request encoding: the first part is json containing the response metadata, and the second part is the request body.  (Unlike requests, there are always exactly 2 parts instead of 1-or-2 parts).

The json part is an object with fields `"code"` containing the HTTP response code and `"headers"` containing HTTP headers (with always-lower-case header names).  For instance:

```json
{"code":200,"headers":{"content-type":"application/json"}}
```
and
```json
{"id":789}
```
encoded as:
```
l58:{"code":200,"headers":{"content-type":"application/json"}}10:{"id":789}e
```
or, for an attachment download (where the response would not be json):
```json
{"code":200,"headers":{"content-type":"application/octet-stream","content-disposition":"attachment;filename*=UTF-8''myfile.txt"}}
```
with body `hello world`:
```
l129:{"code":200,"headers":{"content-type":"application/octet-stream","content-disposition":"attachment;filename*=UTF-8''myfile.txt"}}11:hello worlde
```

### DIFF
--- a/tests/test_onion_requests.py
+++ b/tests/test_onion_requests.py
@@ -1,0 +1,289 @@
+from sogs.web import app
+from sogs import crypto
+from sogs.hashing import blake2b
+from sogs import utils
+from nacl.bindings import (
+    crypto_scalarmult,
+    crypto_aead_xchacha20poly1305_ietf_encrypt,
+    crypto_aead_xchacha20poly1305_ietf_decrypt,
+)
+from Cryptodome.Cipher import AES
+import nacl.utils
+import struct
+import json
+import auth
+
+from nacl.public import PrivateKey
+
+# ephemeral X25519 keypair for use in tests
+a = PrivateKey(bytes.fromhex('ec32fb5766cf52b1b5d7b0bff08e29f5c0c58ca19beaf6a5c7d3dd8ac7ced963'))
+A = a.public_key
+assert A.encode().hex() == 'd79a50b82ba8ca665f854382b42ba159efd16eef87409e97a8d07395b9492928'
+
+# For xchacha20 with use the libsodium recommended shared key of H(aB || A || B), where H(.) is
+# 32-byte Blake2B
+shared_xchacha20_key = blake2b(
+    (
+        crypto_scalarmult(a.encode(), crypto.server_pubkey_bytes),
+        A.encode(),
+        crypto.server_pubkey_bytes,
+    )
+)
+
+# AES-GCM onion requests were implemented using the somewhat weaker shared key of just aB:
+shared_aes_key = crypto_scalarmult(a.encode(), crypto.server_pubkey_bytes)
+
+
+def build_payload(inner_json, inner_body=None, *, v, enc_type, outer_json_extra={}):
+    """Encrypt and encode a payload for sogs"""
+
+    if not isinstance(inner_json, bytes):
+        inner_json = json.dumps(inner_json).encode()
+
+    if v == 3:
+        assert inner_body is None
+        inner_data = inner_json
+    elif v == 4:
+        inner_data = b''.join(
+            (
+                b'l',
+                str(len(inner_json)).encode(),
+                b':',
+                inner_json,
+                *(() if inner_body is None else (str(len(inner_body)).encode(), b':', inner_body)),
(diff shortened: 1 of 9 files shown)


tests/test_onion_requests.py
```diff
@@ -0,0 +1,289 @@
...
+# ephemeral X25519 keypair for use in tests
+a = PrivateKey(bytes.fromhex('ec32fb5766cf52b1b5d7b0bff08e29f5c0c58ca19beaf6a5c7d3dd8ac7ced963'))
+A = a.public_key
+assert A.encode().hex() == 'd79a50b82ba8ca665f854382b42ba159efd16eef87409e97a8d07395b9492928'
+
+# For xchacha20 with use the libsodium recommended shared key of H(aB || A || B), where H(.) is
+# 32-byte Blake2B
+shared_xchacha20_key = blake2b(
+    (
+        crypto_scalarmult(a.encode(), crypto.server_pubkey_bytes),
```
majestrate, line 27: `crypto.server_pubkey_hash_bytes` ?

jagerman: Maybe I'm not understanding.  We're computing `H(aB || A || B)` here, and the server can compute the same value via `H(bA || A || B)`

majestrate (approved):

PR #56, opened 2022-02-14
Room accessible flag & DB migration refactor

jagerman:
1. `sogs.db` was getting messy with all the migration code; this moves it all into multiple files in `sogs.migrations.*`, where each migration submodule has a `migrate(conn)` function to perform a migration.

   Aside from moving all the migrations, this also renames/updates the existing `sogs.migrate01x` to `sogs.migrations.v_0_1_x` for consistency.

2.  This adds a new 'accessible' permission bit that allows restricting all metadata room info.  This allows you to distinguish between rooms that should be entirely secret (i.e. that random users can't learn anything about):

        read=False, accessible=False

    and rooms where users can get metadata but not read messages:

        read=False, accessible=True


sogs/migrations/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+    ):
+        changes = False
+        with db.transaction(conn):
+            changes = migration.migrate(conn)
+        if changes:
+            db.metadata.clear()
+            db.metadata.reflect(bind=db.engine, views=True)
+
+
+#        migrate_v01x,
```
majestrate: what are these comments for?

sogs/migrations/import_hacks.py
```diff
@@ -0,0 +1,39 @@
...
+            db.HAVE_FILE_ID_HACKS = True
+
+    try:
+        rows = conn.execute(
+            "SELECT room, old_message_id_max, message_id_offset FROM room_import_hacks"
+        )
+        for (room, id_max, offset) in rows:
+            db.ROOM_IMPORT_HACKS[room] = (id_max, offset)
+    except Exception:
+        pass
```
majestrate, line 39: do we really want a silent fail here?

sogs/migrations/new_columns.py
```diff
@@ -0,0 +1,29 @@
...
+
+def migrate(conn):
+    """
+    New columns that might need to be added that don't require more complex migrations beyond simply
+    adding the column.
+    """
+
+    from .. import db
+
+    new_table_cols = {
```
majestrate, line 12: why a set instead of a list?

sogs/model/room.py
```diff
@@ -1400,15 +1447,15 @@ def get_rooms_with_permission(
...
-def get_readable_rooms(user: Optional[User] = None):
+def get_accessible_rooms(user: Optional[User] = None):
     """
-    Get a list of rooms that a user can access; if user is None then return all publicly readable
+    Get a list of rooms that a user can access; if user is None then return all publicly accessible
     rooms.
     """
     if user is None:
-        result = query("SELECT * FROM rooms WHERE read ORDER BY token")
+        result = query("SELECT * FROM rooms WHERE (read OR accessible) ORDER BY token")
```
majestrate, line 1456: should this be `read AND accessible` ?

sogs/routes/auth.py
```diff
@@ -135,6 +135,36 @@ def require_mod(room, *, admin=False):
...
 
+def accessible_required(f):
+    """Decorator for an endpoint that requires a user have accessible or read permission in the
+    given room.  The function must take a `room` argument by name, as is typically used with flask
+    endpoints with a `<Room:room>` argument."""
+
+    @wraps(f)
+    def required_accessible_wrapper(*args, room, **kwargs):
+        if not room.check_accessible(g.user):
+            abort_with_reason(http.FORBIDDEN, "This room is not accessible", warn=False)
```
majestrate: this can leak info between rooms not existing and rooms not being accessible.

sogs/model/room.py
jagerman: I don't think so.  `accessible` only has meaning when `read` is false; when `read` is true we don't care what `accessible` is (or to put it another way: `read` implies a room being accessible).  So a room is accessible to you if: `(read or (not read and accessible))` which simplifies to `read or accessible`.

sogs/routes/auth.py
jagerman: Yes, but does that actually matter?

sogs/migrations/__init__.py
jagerman: Oops.  Leftover scaffolding from when I was moving things.  Removed.

sogs/migrations/import_hacks.py
jagerman: Yes: if the table doesn't exist the query fails and throws, which is fine: we have no hack values to worry about.

sogs/migrations/new_columns.py
jagerman: Seems a bit cleaner to declare, and we don't care about the order these get applied.

sogs/routes/auth.py
majestrate: yes, the point of this flag was to prevent this type of info leak.

jagerman: Updated to 404 instead.

majestrate (approved):

PR #69, opened 2022-02-25
file upload endpoints

### LINKED ISSUE #65: File upload endpoints
Three things here:
- Implementing new endpoints so submit files, which will give them a default 1h expiry time, and take a suggested filename.
- Download endpoints (*not* using JSON), along with filename, etc. provided in headers.
- Add recognition of "files" parameter in the post and edit message endpoints, so that it associates any listed files that don't already have an associated posts with the new (or edited) post, *and* updates the expiry on the files to the server default expiry time.

majestrate:
add room file upload and room file fetching endpoints with unit tests

Fixes #65 #80


sogs/routes/rooms.py
```diff
@@ -264,3 +269,39 @@ def message_unpin(room, msg_id):
...
+def upload_file(room):
+    """ upload a file to a room """
+
+    if not room.check_upload(g.user):
+        abort(http.FORBIDDEN)
+
+    filename = request.headers.get("X-Filename")
+    if not filename:
+        abort(http.BAD_REQUEST)
+    id = room.upload_file(request.data, g.user, filename=secure_filename(filename))
```
jagerman: Don't need `secure_filename` here: `room.upload_file` already takes care of sanitizing it.

sogs/routes/rooms.py
```diff
@@ -264,3 +269,39 @@ def message_unpin(room, msg_id):
...
+@auth.user_required
+def serve_file(room, fileId, filename):
+    """
+    serves a file uploaded to a room
+    """
+    room_file = room.get_file(fileId)
+    if not room_file:
+        abort(http.NOT_FOUND)
+
+    resp = make_response(send_from_directory(abspath(curdir), room_file.path, as_attachment=True))
```
jagerman: Switch to `send_file` instead so that you can specify the path and `download_filename` separately, because basename(path) != filename: the stored filename has a prefix of `{id}_`, and will be shortened if the given filename was long.  (We don't need send_from_directory's path protections because `path` is sanitized before we insert it into the database).

sogs/routes/rooms.py
```diff
@@ -264,3 +269,39 @@ def message_unpin(room, msg_id):
...
+
+    resp.headers["Date"] = http_date(room_file.uploaded)
+    resp.headers["Expires"] = http_date(room_file.expiry)
+    resp.headers["Content-Type"] = 'application/octet-stream'
```
jagerman: Should we attempt to guess the content-type when we have a filename with a known extension?  Presumably there is some Python module that makes that easy.

tests/request.py
```diff
@@ -10,16 +10,24 @@ def sogs_get(client, url, user):
     return client.get(url, headers=x_sogs_for(user, "GET", url))
 
 
+def sogs_post_raw(client, url, data, user, *, ctype='application/octet-stream', extra_headers=None):
```
jagerman:
```suggestion
def sogs_post_raw(client, url, data, user, *, ctype='application/octet-stream', extra_headers={}):
```

tests/request.py
```diff
@@ -10,16 +10,24 @@ def sogs_get(client, url, user):
...
+    """
+    POSTs a test `client` request to `url` with the given `data` as bytes body and X-SOGS-*
+    signature headers signing the request for `user`.
+    """
+    headers = x_sogs_for(user, "POST", url, data)
+    for k in extra_headers or []:
+        if k not in headers:
+            headers[k] = extra_headers[k]
+    return client.post(url, data=data, content_type=ctype, headers=headers)
```
jagerman:
```suggestion
    return client.post(url, data=data, content_type=ctype, headers={**headers, **extra_headers})
```
Simplification, and also I don't think the `not in headers` makes sense here: if I give an extra header value that conflicts with one of the X-SOGS headers then I'd expect it to override when the header already exists rather than silently being ignored.

tests/test_room_routes.py
```diff
@@ -6,7 +6,8 @@
 import sogs.config
 import werkzeug.exceptions as wexc
 from util import pad64
-from request import sogs_get, sogs_post, sogs_put
+from request import sogs_get, sogs_post, sogs_put, sogs_post_raw
+import secrets
```
jagerman:
```suggestion
from nacl.utils import random
```
Not a big deal, just a minor preference for libsodium's `randombuf_bytes`, partly because `token_bytes` is a stupid name.

tests/test_room_routes.py
```diff
@@ -1041,3 +1042,42 @@ def test_edits(client, room, user, user2, mod, global_admin):
     url_get = "/room/test-room/messages/since/4"
     r = sogs_get(client, url_get, user)
     assert r.json == []
+
+
+def test_file_upload(client, room, user):
+
+    filename = 'testfile.bin'
+    url_post = f"/room/{room.token}/file"
+    file_content = secrets.token_bytes(1024)
```
jagerman:
```suggestion
    file_content = random(1024)
```

tests/test_room_routes.py
```diff
@@ -1041,3 +1042,42 @@ def test_edits(client, room, user, user2, mod, global_admin):
...
+    assert r.data == file_content
+
+
+def test_file_upload_banned_user(client, room, banned_user):
+    filename = 'testfile.bin'
+    url_post = f"/room/{room.token}/file"
+    r = sogs_post_raw(
+        client,
+        url_post,
+        secrets.token_bytes(1024),
```
jagerman:
```suggestion
        random(1024),
```

tests/test_room_routes.py
```diff
@@ -1041,3 +1042,42 @@ def test_edits(client, room, user, user2, mod, global_admin):
...
+    r = sogs_post_raw(client, url_post, file_content, user, extra_headers={"X-filename": filename})
+    assert r.status_code == 201
+    assert 'id' in r.json
+    id = r.json.get('id')
+    assert id is not None
+    assert id != 0
+
+    r = sogs_get(client, f'/room/test-room/file/{id}/{filename}', user)
+    assert r.status_code == 200
+    assert r.data == file_content
```
jagerman, line 1267: Some other things to add to the tests:

- test that the file exists on disk in the right place (`uploads/test-room/1_{filename}`)
- for a simple filename check that we are getting `Content-Disposition` set to `attachment; filename="testfile.bin"`.  (Or possibly flask is producing `attachment; filename*=UTF-8''testfile.bin`; if so, use that for the test instead).
- upload with unicode characters in the filename (currently we strip that out of the actual path/filename we write to disk (which is good), but I think we might also be stripping it out of the `filename`, which we shouldn't).  E.g. `my 🎂.txt` should end up as (I think) `uploads/test-room/2_my_.txt` for path, `my 🎂.txt` for filename, and Content-Disposition on the download should be: `attachment; filename*=UTF-8''my%20%F0%9F%8E%82.txt`.
- if a user has `upload=false` set they shouldn't be able to upload an attachment.
- if a user has `write=false` permission (even with `upload` true) they shouldn't be able to upload an attachment.
- a user with `read=false` access should not be able to download an attachment.
- make an upload, pin it to a room (which will set the upload expiry to 0), and then verify that downloading it doesn't include an Expires header.

sogs/routes/rooms.py
```diff
@@ -264,3 +269,39 @@ def message_unpin(room, msg_id):
...
+
+@rooms.post("/room/<Room:room>/file")
+@auth.user_required
+def upload_file(room):
+    """ upload a file to a room """
+
+    if not room.check_upload(g.user):
+        abort(http.FORBIDDEN)
+
+    filename = request.headers.get("X-Filename")
```
jagerman: This header isn't going to work for unicode filenames; the only standardized encoding of unicode through HTTP headers (AFAIK) is [RFC 5987](https://www.rfc-editor.org/rfc/rfc5987), which requires something like:

```
Foo: v*=UTF-8''xxxxxx
```
where xxxxx is the url-encoded utf-8 byte sequence, for arbitrary `Foo` and `v`.  E.g. in a Content-Disposition header for `my 🎂.txt`:
```
Content-Disposition: attachment; filename*=UTF-8''my%20%F0%9F%8E%82
```

We could do something similar for the `X-Filename` header, but it seems like it'd be just as easy to use the `Content-Disposition` header as above for the upload request rather than `X-Filename`.

(And hopefully werkzeug exposes some means of parsing such unicode-tagged headers).

sogs/routes/rooms.py
```diff
@@ -264,3 +269,39 @@ def message_unpin(room, msg_id):
...
+    serves a file uploaded to a room
+    """
+    room_file = room.get_file(fileId)
+    if not room_file:
+        abort(http.NOT_FOUND)
+
+    resp = make_response(send_from_directory(abspath(curdir), room_file.path, as_attachment=True))
+
+    resp.headers["Date"] = http_date(room_file.uploaded)
+    resp.headers["Expires"] = http_date(room_file.expiry)
```
jagerman: What happens here if `expiry` is None?

sogs/routes/rooms.py
```diff
@@ -264,3 +269,39 @@ def message_unpin(room, msg_id):
...
+    if not filename:
+        abort(http.BAD_REQUEST)
+    id = room.upload_file(request.data, g.user, filename=secure_filename(filename))
+    resp = make_response(jsonify({"id": id}))
+    resp.status_code = http.CREATED
+    return resp
+
+
+@rooms.get("/room/<Room:room>/file/<int:fileId>/<filename>")
+@auth.user_required
```
```suggestion
@auth.read_required
```

jagerman (changes requested):
Left various comments above.

One more piece is needed here, though, to avoid orphan uploads by associating uploads with posts in sogs: see #65.

sogs/routes/rooms.py
majestrate: will that ever happen?

jagerman: Yup: room images have a null expiry, as will attachments in pinned posts (though that isn't implemented yet).

majestrate: what should we send in this case, expiry is 0?

jagerman: I think just omit the Expiry header.

jagerman: From what I gather, there is no "never" value for Expiry headers, and a browser is supposed to treat anything that it fails to parse as 0 which it is supposed to mean "it expired a long time ago"

sogs/routes/rooms.py
```diff
@@ -301,7 +301,7 @@ def serve_file(room, fileId, filename):
     resp = make_response(send_from_directory(abspath(curdir), room_file.path, as_attachment=True))
 
     resp.headers["Date"] = http_date(room_file.uploaded)
-    resp.headers["Expires"] = http_date(room_file.expiry)
-    resp.headers["Content-Type"] = 'application/octet-stream'
+    if room_file.expiry and room_file.expiry > 0:
```
jagerman: 0 is a special value that means "expire now", which isn't something we should really worry about here.  I think, however, that `model.room.Room.get_file` should be taking care of not returning expired files that haven't hit the cleanup job yet, i.e. by adding a not expired check here:
```Python
if row:
    return File(row)
```
i.e.:
```Python
if row and row['expiry'] > time.time():
    return File(row)
```

jagerman: (It'd be even better if it was in the query, except then we'd potentially trigger the extra file_id_hacks query crap unnecessarily).

sogs/routes/rooms.py
```diff
@@ -296,11 +296,13 @@ def serve_file(room, fileId, filename):
     room_file = room.get_file(fileId)
     if not room_file:
         abort(http.NOT_FOUND)
+    if room_file.expiry == 0:
+        abort(http.NOT_FOUND)
```
jagerman: See my other comments: I think `model.Room` should be responsible for not returning an expired File rather than routes.

jagerman:
One other thing: we should clean up the `upload/test-room` directory as part of the tests DB cleanup (would probably work at the end of the `db` fixture in `conftest.py`).

sogs/routes/rooms.py
```diff
@@ -337,3 +339,53 @@ def poll_room_info(room, info_updated):
             result['global_admin'] = True
 
     return jsonify(result)
+
+
+@rooms.post("/room/<Room:room>/file")
+@auth.user_required
+def upload_file(room):
+    """ upload a file to a room """
```
jagerman: Needs to be rebased on dev and the docs moved out of api.yaml to here (updated if needed).

sogs/routes/rooms.py
```diff
@@ -337,3 +339,53 @@ def poll_room_info(room, info_updated):
...
+    resp.status_code = http.CREATED
+    return resp
+
+
+@rooms.get("/room/<Room:room>/file/<int:fileId>/<filename>")
+@auth.read_required
+def serve_file(room, fileId, filename):
+    """
+    serves a file uploaded to a room
+    """
```
jagerman, line 500: Move docs from api.yaml

jagerman:
#77 should be ready for merge or cherry-picking now, which adds the `files` arguments to add_post/edit_post to handle updating file ownership.

jagerman (approved):

PR #70, opened 2022-02-25
clean up room viewer urls

majestrate:
seperate room viewer routes into its own file

### DIFF
--- a/sogs/utils.py
+++ b/sogs/utils.py
@@ -102,7 +102,7 @@ def bencode_consume_string(body: memoryview) -> Tuple[memoryview, memoryview]:
 
 
 def server_url(room):
-    return '{}/{}?public_key={}'.format(config.URL_BASE, room or '', crypto.server_pubkey_hex)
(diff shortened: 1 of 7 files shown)


sogs/utils.py
```diff
@@ -102,7 +102,7 @@ def bencode_consume_string(body: memoryview) -> Tuple[memoryview, memoryview]:
 
 
 def server_url(room):
-    return '{}/{}?public_key={}'.format(config.URL_BASE, room or '', crypto.server_pubkey_hex)
+    return '{}/r/{}?public_key={}'.format(config.URL_BASE, room or '', crypto.server_pubkey_hex)
```
jagerman: This is going to have to be delayed until it gets supported by Session, unfortunately.

jagerman: Either revert this (and we'll apply it later, once Session is ready), or perhaps make a `config.` global for whether or not we add the `/r` here.

jagerman (approved):
One tweak to not deploy the `/r` yet in the open group URL, other than that looks good.

PR #72, opened 2022-02-26
Add import CI tests & various db fixes

### LINKED ISSUE #62: Room deletion fails
Currently deleting a room with at least some active users and/or posts fails with a foreign key constraint violation.

This makes me suspect one or more of the foreign keys are not declared `ON DELETE CASCADE`.

jagerman:
This adds a CI job and scripts that take a v0.1.10 test database, import it, and test that we get the expected result.  It currently includes three tests: 0.1.10 -> curr (for sqlite + postgresql), and 0.1.10 -> 0.2.0 -> curr (for sqlite).

This found all sorts of upgrade issues, fixed herein.

- v0.1.x import has been non-functional for a long time (since just after the 0.2.0 tag)
- make v0.1.x import work on postgresql
- inbox was only in the migration code but missing from the master schemas
- the import hacks table had non-cascading foreign keys, which prevented deletion; this PR fixes them and adds a migration fix for existing setups (fixes #62)
- message_metadata upgrades broke seqno column renames under sqlite, and ended up with the wrong name under postgres
- sqlite inbox had incorrect 1d default expiry instead of 15d
- `citext` postgresql extension was getting put in the sogs_test schema by the postgresql test suite, which made it unusable in the public schema if you used the same db for something other that pytest.  Fixed it to always put it in the public schema.

~~Still TODO:~~
- [x] create from-scratch v0.2.0 snapshot (i.e. initial non-sogs install) for sqlite3
- [x] create from-initial-pg snapshot upgrade for postgresql


sogs/migrations/v_0_1_x.py
```diff
@@ -27,6 +27,14 @@ def migrate(conn):
     return True
 
 
+def sqlite_connect_readonly(path):
+    import sqlite3
+
+    conn = sqlite3.connect(f"file:{path}?mode=ro", uri=True, isolation_level=None)
+    conn.row_factory = sqlite3.Row
+    return conn
```
majestrate, line 35: since this is used in a `with` statement does returning here work instead of using a `yield`?

majestrate (commented):
nitpick: the urls from oxen.rocks should probably be in an /assets/ sub dir

jagerman:
> nitpick: the urls from oxen.rocks should probably be in an /assets/ sub dir

Done.

majestrate (approved):

PR #74, opened 2022-02-28
add endpoints for deleting posts from a room

majestrate:


sogs/routes/messages.py
```diff
@@ -307,6 +307,34 @@ def edit_message(room, msg_id):
...
+
+    - 403 Forbidden — returned if the invoking user does not have admin permission in this room.
+
+    - 404 Not Found — returned if the given post was not found in this room.
+
+    """
+    try:
+        room.delete_posts([msg_id], deleter=g.user)
+    except model.exc.BadPermission:
+        abort(http.FORBIDDEN)
```
jagerman, line 332: Don't need to catch this: we have an error handler in sogs.model.exc that does this BadPermission -> abort(Forbidden) already.

jagerman:
Other than that, looks good.

jagerman (approved):

PR #75, opened 2022-03-01
outbox endpoints

majestrate:
add outbox endpoints, refactor common logic for in/out boxes into common helper function


sogs/routes/users.py
```diff
@@ -76,16 +76,89 @@ def extract_rooms_or_global(req, admin=True):
...
         "expires_at": msg.expires_at,
         "message": utils.encode_base64(msg.data),
-        "sender": msg.sender.session_id,
+        key: getattr(msg, key).session_id,
     }
 
 
+def _box(func, **kwargs):
+    """ handle inbox/outbox endpoints common logic """
+    if g.user.banned or not g.user.is_blinded:
```
jagerman: Don't need this check on `g.user.banned`: if the user is globally banned the request gets rejected during authentication.

sogs/routes/users.py
```diff
@@ -76,16 +76,89 @@ def extract_rooms_or_global(req, admin=True):
...
+        abort(http.FORBIDDEN)
+    keys = {'to': 'sender', 'sent': 'recipient'}
+    key = keys.get(func)
+    limit = utils.get_int_param('limit', 100, min=1, max=256, truncate=True)
+    get = getattr(Message, func)
+    msgs = [_serialize_message(msg, key=key) for msg in get(user=g.user, limit=limit, **kwargs)]
+    if msgs or not kwargs:
+        return jsonify(msgs)
+    return Response('', status=http.NOT_MODIFIED)
```
jagerman: This indirection with `func` and key as strings that goes through a dict lookup makes this a bit harder to follow than necessary.  If this were just a bool it would be clearer:

```Python
def _box(in:bool=True, **kwargs):
    # ...
    get = Message.to if in else Message.sent
    msgs = [_serialize_message(msg) for msg in get(user=g.user, limit=limit, **kwargs)]
```

sogs/routes/users.py
```diff
@@ -76,16 +76,89 @@ def extract_rooms_or_global(req, admin=True):
...
-def _serialize_message(msg):
+def _serialize_message(msg, *, key):
     return {
         "id": msg.id,
         "posted_at": msg.posted_at,
         "expires_at": msg.expires_at,
         "message": utils.encode_base64(msg.data),
-        "sender": msg.sender.session_id,
+        key: getattr(msg, key).session_id,
     }
```
jagerman: I think we should just include both sender and recipient for both inbox/outbox messages, since it slightly simplifies the logic here and unifies the output.

majestrate: yea

sogs/routes/users.py
```diff
@@ -76,16 +76,89 @@ def extract_rooms_or_global(req, admin=True):
...
 
 
+def _box(func, **kwargs):
```
jagerman: Doesn't make sense to me to use a `kwargs` that can only ever have one argument (`since`) rather than just taking `*, since=None` as arguments.

sogs/routes/users.py
```diff
@@ -76,16 +76,89 @@ def extract_rooms_or_global(req, admin=True):
...
+
+    - `id` — the unique integer message id.
+    - `posted_at` — unix timestamp (float) when the message was received by SOGS.
+    - `expires_at` — unix timestamp (float) when SOGS will expire and delete the message.
+    - `message` — the encrypted message body.
+    - `recipient` — the (blinded) Session ID of the recipient of the message.
+
+    # Encryption details
+
+    see [/inbox](#GET-inbox)
```
jagerman:
```suggestion
    See [/inbox](#GET-inbox)
```

sogs/routes/users.py
```diff
@@ -76,16 +76,89 @@ def extract_rooms_or_global(req, admin=True):
...
+    # Return value
+
+    This method, on success, returns *either* a 200 (OK) status code with a list of 1 or more new
+    messages, or else returns a 304 (Not Modified) response with an empty body to indicate that
+    there are no messages for this user newer than the given ID.
+
+    If there are messages this returns a JSON array of up to `limit` (default 100) messages, with
+    oldest messages first.  Each element is exactly as described in the [all messages](#GET-outbox)
+    endpoint.
+    """
```
jagerman: I feel like we can substantially dry out these docstrings by extracting all the common bits into globals and using f-strings for the docstring.

jagerman: (Just make sure the global has the same 4-space indent, so that it gets properly stripped off when the docstring gets cleaned up).

majestrate: dry'd up a bit

jagerman (approved):

PR #77, opened 2022-03-02
Add files.message foreign key and related code

jagerman:
- Adds triggers to expire attachments when their post is delete.
- Adds triggers so that pinning/unpinning automatically makes any associated attachments unexpiring/expiring.
- Adds model.Room functions to associate new files with a post when adding/editing a post.
- Adds a test stub for PR #69


sogs/model/room.py
```diff
@@ -1363,7 +1399,27 @@ def unpin_all(self, admin: User):
             app.logger.warning("Unable to unpin all messages from {self}: {admin} is not an admin")
             raise BadPermission()
 
-        count = query("DELETE FROM pinned_messages WHERE room = :r", r=self.id).rowcount
+        with db.transaction():
+            if config.UPLOAD_DEFAULT_EXPIRY_DAYS != 15:
```
majestrate: why are we checking this in this manner?

jagerman: It was an optimization, but a premature one after thinking about it, so removed.

majestrate (approved):

PR #82, opened 2022-03-08
Add Remove all posts for user from room endpoint

### LINKED ISSUE #67: Delete all user's posts endpoint
We're missing an implementation of the delete-all-a-users-posts endpoint, `/user/{sessionId}/deleteMessages` (which should have the option to either delete from specific rooms, or from all rooms where the caller is a moderator -- very similarly to how `/user/{sessionId}/ban` works now).

majestrate:
* Unit tests
* Fix issue with Room.delete_all_posts where exception thrown when no files uploaded

fixes #67

### DIFF
--- a/sogs/routes/rooms.py
+++ b/sogs/routes/rooms.py
@@ -1,5 +1,5 @@
 from .. import config, db, http
-from ..model import room as mroom, exc
+from ..model import room as mroom, exc, user as muser
 from ..web import app
 from . import auth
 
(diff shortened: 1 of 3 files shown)


sogs/routes/rooms.py
```diff
@@ -1,5 +1,6 @@
 from .. import config, db, http
 from ..model import room as mroom
+from ..model import user as muser
```
```suggestion
from ..model import room as mroom, user as muser
```

jagerman (approved):

PR #86, opened 2022-03-17
room ban info endpoint

### LINKED ISSUE #63: Add endpoints for listing bans
In old SOGS there was a `/block_list` that just returns the list of all banned session IDs.  This doesn't work well for new SOGS because we have supplemental info, such as a possible ban timeout, as well as different types of restrictions (for example, someone might only be muted but not entirely banned).

We can, of course, just return a dump of the ban info, but that isn't great either because a ban list almost never shrinks, which means the returned data is going to be growing forever.

I think, instead, we need some sort of incremental ban mechanism so that moderators can query bans/permissions/etc. and receive updates to that list without needing to download the whole list every time.

majestrate:
Fetch room permissions routes. Basic unit tests and basic functionality.

Fixes #63

### DIFF
--- a/sogs/model/room.py
+++ b/sogs/model/room.py
@@ -1517,8 +1517,6 @@ class Room:
         ):
             data = dict()
             for k in row.keys():
-                if k in ('banned', 'moderator', 'admin', 'visible_mod') and not row[k]:
-                    continue  # only include banned when true
                 if row[k] is not None and k not in ('session_id', 'room', 'user'):
                     data[k] = bool(row[k])
             ret[row['session_id']] = data
--- a/sogs/model/room.py
+++ b/sogs/model/room.py
@@ -1511,22 +1511,22 @@ class Room:
         if not self.check_moderator(mod):
             app.logger.warning("unable to get room permissions for user")
             raise BadPermission()
-        with db.transaction():
-            ret = dict()
-            for row in query(
-                """SELECT session_id, upo.* FROM user_permission_overrides upo
-                JOIN users ON "user" = users.id WHERE room = :r""",
-                r=self.id,
-            ):
-                data = dict()
-                for k in row.keys():
-                    if k not in ('session_id', 'room', 'user'):
-                        if k in ('banned', 'moderator', 'admin', 'visible_mod') and not row[k]:
-                            continue  # only include banned when true
-                        if row[k] is not None:
-                            data[k] = bool(row[k])
-                ret[row['session_id']] = data
-            return ret
+
+        ret = dict()
+        for row in query(
+            """SELECT session_id, upo.* FROM user_permission_overrides upo
+            JOIN users ON "user" = users.id WHERE room = :r""",
+            r=self.id,
+        ):
+            data = dict()
+            for k in row.keys():
+                if k not in ('session_id', 'room', 'user'):
+                    if k in ('banned', 'moderator', 'admin', 'visible_mod') and not row[k]:
+                        continue  # only include banned when true
+                if row[k] is not None:
+                    data[k] = bool(row[k])
+            ret[row['session_id']] = data
+        return ret
 
 
 def get_rooms():
--- a/sogs/model/room.py
+++ b/sogs/model/room.py
@@ -1517,10 +1517,9 @@ class Room:
         ):
             data = dict()
             for k in row.keys():
-                if k not in ('session_id', 'room', 'user'):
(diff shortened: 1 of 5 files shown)


sogs/model/room.py
```diff
@@ -1464,8 +1464,34 @@ def unpin(self, msg_id: int, admin: User):
...
+        with db.transaction():
+            ret = dict()
+            for row in query(
+                """SELECT session_id, upo.* FROM user_permission_overrides upo
+                JOIN users ON "user" = users.id WHERE room = :r""",
+                r=self.id,
+            ):
+                data = dict()
+                for k in row.keys():
+                    if k not in ('session_id', 'room', 'user'):
```
jagerman:
```suggestion
                    if k not in ('session_id', 'room', 'user'):
                        if k == 'banned' and not row[k]:
                            continue  # only include banned when true
```

sogs/routes/rooms.py
```diff
@@ -278,6 +278,24 @@ def update_room(room):
...
+    # Query Parameters
+
+    TODO: implement me
+
+    # Return Value
+
+    dict of session_id to current permission info
+
+    """
+    return jsonify(room.export_permissions(g.user))
```
jagerman: I'm worried this might get big, as is, because it'll be something like:
```json
{"15abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789": {"read": true, "upload": false},"15bcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789a": {"banned": true},
...}
```

It might be worth encoding it to something smaller, perhaps by instead of a dict of "perm": true|false, we could return a set (I dunno if jsonify is happy converting a set to a list; might need to be done manually) containing:
"r"/"-r" - included if read is not null
"w"/"-w" included if write is not null
"a"/"-a" included if accessible is not null
"u"/"-u" included if upload is not null

Some of the flags are non-nullable (because they are additive rather than overriding) and so wouldn't have "-" variations, i.e. would only be included when true:
"b" would be included only if banned is true
"A" would be included only if admin is true
"M" would be included only if moderator is true *and* admin is not true
"V" would be included only if moderator is true and visible_mod is true

sogs/model/room.py
```diff
@@ -1464,8 +1464,36 @@ def unpin(self, msg_id: int, admin: User):
...
+                        if k == 'banned' and not row[k]:
```
jagerman: My bad, there are actually a few of these non-nullables:
```suggestion
                        if k in ('banned', 'moderator', 'admin', 'visible_mod') and not row[k]:
```

sogs/model/room.py
```diff
@@ -1464,8 +1464,36 @@ def unpin(self, msg_id: int, admin: User):
...
 
+    def export_permissions(self, mod):
+        """
+        export room permissions in full,
+        returns a dict of session_id -> permissions dict (a dict of permission type to bool)
+        """
+        if not self.check_moderator(mod):
+            app.logger.warning("unable to get room permissions for user")
+            raise BadPermission()
+        with db.transaction():
```
jagerman: There's only a single query here, no need for a transaction.

sogs/model/room.py
```diff
@@ -1494,6 +1491,33 @@ def export_permissions(self, mod):
...
+            r=self.id,
+        ):
+            data = dict()
+            for k in row.keys():
+                if k in ('room', 'user'):
+                    continue
+                if k in ('at', 'session_id'):
+                    data[k] = row[k]
+                else:
+                    data[k] = bool(row[k])
```
jagerman, line 1548: This needs to preserve `None`s as `None`

sogs/routes/rooms.py
```diff
@@ -278,6 +278,61 @@ def update_room(room):
...
+    # Return Value
+
+    dict of session_id to current permissions,
+    if compact is set to 1 this will be a string,
+    otherwise it will be a dict containing the name of the permission mapped to a boolean value.
+    """
+    transform = lambda x: x  # noqa: E731
+    if request.args.get("compact", type=int, default=0):
+        transform = compact_permissions
+    return jsonify(transform(room.permissions))
```
```suggestion
    perm = room.permissions
    if request.args.get("compact", type=int, default=0):
        perm = compact_permissions(perm)
    return jsonify(perm)
```

sogs/model/room.py
jagerman: Never mind, misread the diff.

jagerman (approved):

PR #88, opened 2022-03-21
add instructions to build a quick docker image for dev and production

Bilb:
What's left to be done is 
- make sure the container restarts if it crashes (unsure if we can do this)
- for the production one, add a way to keep the sogs command running while the container is up 
- deploy to dockerhub?


tests/util.py
```diff
@@ -15,7 +15,7 @@ def pad64(data: Union[bytes, str]):
 
 
 class FuzzyTime:
-    epsilon = 0.25
+    epsilon = 1
```
majestrate: why?

Bilb: removed

docker/DockerFile.dev
```diff
@@ -0,0 +1,26 @@
+# check docker/Readme.md
+
+FROM ubuntu:impish
```
jagerman:
```suggestion
FROM debian:stable
```
Impish will fall out of supported status in a couple months.

docker/DockerFile.dev
```diff
@@ -0,0 +1,26 @@
...
+RUN apt update
+RUN apt upgrade -y
+
+RUN apt-get install -y locales && locale-gen en_US.UTF-8
+ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
+
+RUN DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt -y install curl build-essential libssl-dev pkg-config lsb-release
+RUN curl -so /etc/apt/trusted.gpg.d/oxen.gpg https://deb.oxen.io/pub.gpg
+RUN echo "deb https://deb.oxen.io $(lsb_release -sc) main" | tee /etc/apt/sources.list.d/oxen.list
```
jagerman:
```suggestion
RUN echo "deb https://deb.oxen.io $(lsb_release -sc) main" >/etc/apt/sources.list.d/oxen.list
```

jagerman: `| sudo tee filename` is just a (slight) abuse to `sudo` the file redirection, but not needed here.

docker/DockerFile.dev
```diff
@@ -0,0 +1,26 @@
+# check docker/Readme.md
+
+FROM ubuntu:impish
+RUN apt update
+RUN apt upgrade -y
```
jagerman:
```suggestion
RUN apt-get update
RUN apt-get dist-upgrade -y
```

docker/DockerFile.dev
```diff
@@ -0,0 +1,26 @@
...
+
+RUN apt-get install -y locales && locale-gen en_US.UTF-8
+ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
```
jagerman: Were either of these necessary?  I really think they shouldn't be, and if they are we probably need a fix in SOGS so that they aren't.

docker/DockerFile.dev
```diff
@@ -0,0 +1,26 @@
...
+
+RUN DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt -y install curl build-essential libssl-dev pkg-config lsb-release
```
jagerman:
```suggestion
RUN DEBIAN_FRONTEND=noninteractive apt-get -y install --no-install-recommends lsb-release
```
I don't think any of the other deps there are needed.

docker/DockerFile.dev
```diff
@@ -0,0 +1,26 @@
...
+RUN curl -so /etc/apt/trusted.gpg.d/oxen.gpg https://deb.oxen.io/pub.gpg
```
jagerman: We can avoid the external request here by using the copy in the repo at contrib/deb.oxen.io.gpg.  (I'm not sure how to do that in a docker file; maybe the `COPY` command does this?)

docker/Readme.md
```diff
@@ -0,0 +1,47 @@
+# Docker for pysogs development
+
+The file DockerFile.dev is specifically made for development if you are not running linux.
+It lets you create a docker running linux and the pysogs in it even if you are running Macos or Windows.
+It creates a docker container with the content of this git reposotiry mounted.
```
jagerman:
```suggestion
It creates a docker container with the content of this git repository mounted.
```

docker/Readme.md
```diff
@@ -0,0 +1,47 @@
...
+Basically, whatever you edit in this repository will be represented on the docker container. So when you run the container, it will run your code.
+
```
jagerman:
```suggestion

# Not for production use

This docker image is strictly for development use and not supported for production use.
```

jagerman (changes requested):
Some minor changes included as comments here.

I also think we should move the `/docker` bits into `/contrib/docker` which might make people realize that this isn't for production use.

(There are some community members interested in creating a production docker image for pysogs, though that will look a little different — most likely by using the `python3-sogs` package to install the production version rather than reusing a current dev branch checkout.)

docker/DockerFile.dev
Bilb: I get errors when reading the config file otherwise.
Turns out, I can also force it like this 
```    cp.read(conf_ini, encoding='utf-8')```

File: contrib/docker/DockerFile [hunk 10]
```diff
@@ -0,0 +1,24 @@
...
+RUN echo "deb https://deb.oxen.io $(cat /etc/os-release | grep VERSION_CODENAME | cut -d "=" -f2) main" > /etc/apt/sources.list.d/oxen.list
+RUN cat /etc/apt/sources.list.d/oxen.list
+RUN apt-get update
+
+RUN echo sogs-standalone session-pysogs/url-http string http://example.com | debconf-set-selections
+RUN  apt-get install -y sogs-standalone
+
+# just add an alias to the pysogs start command
+RUN echo '#!/bin/bash\n/usr/bin/uwsgi --uid _loki --ini /etc/sogs/uwsgi-standalone.ini' > /usr/bin/pysogs && \
+    chmod +x /usr/bin/pysogs
```
jagerman: Might want to rename this to something like `start-sogs` or `start-sogs-uwsgi` so that it doesn't conflict with the `sogs` command to do sogs admin things.

docker/DockerFile.dev
Bilb: found the root cause for this.
https://github.com/oxen-io/session-pysogs/pull/88/commits/4b6e6df207f6d2bb07a480a8a8f60d542ec13d0c

contrib/docker/DockerFile
Bilb: renamed to `start-sogs-uwsgi`

PR #95, opened 2022-03-24
Query optimizations

jagerman:
Fetching the moderator list (which we do frequently) was a very slow query, taking around 15ms on our official SOGS, and is called frequently (it's in the legacy compact_poll results that all Session clients currently use when polling).  This optimizes the query to run make it fast (well under 1ms), and lowered overall postgresql CPU usage on the server by about 80%.  (The query speed gain on sqlite performance is comparable, so CPU reduction should be similar).


tests/test_rooms.py
```diff
@@ -602,3 +602,28 @@ def test_pinning(room, room2, user, mod, admin, global_admin, no_rate_limit):
         room2.pin(7, global_admin)
 
     assert not room2.pinned_messages
+
+
+def test_active_users(room, user, user2):
```
majestrate, line 607: i feel like the usage of cleanup and _refresh can be done with some kind of context manager here.

majestrate (approved):

tests/test_rooms.py
jagerman: Is it actually worth bothering, just for test code like this?

majestrate: no

PR #104, opened 2022-05-09
Avoid 'mentioned' users having their Session id exposed via room page.

ianmacd:
Currently, an @-mention causes the Session id of the mentioned user to be exposed via the room page, viewable in a browser if the default of `http_show_recent = yes` has not been disabled in `/etc/sogs/sogs.ini`.

This allows a would-be attacker to @-mention his victim in an open group, and then read his Session id using a browser.

This patch replaces instances of that Session id with `@hidden`:

image

### DIFF
--- a/sogs/static/view_room.js
+++ b/sogs/static/view_room.js
@@ -53,7 +53,7 @@ const setup = async () => {
                     }
 
                     const plain = Message.decode(data).dataMessage;
-                    e.appendChild(document.createTextNode(plain.profile.displayName +": "+plain.body));
+                    e.appendChild(document.createTextNode(plain.profile.displayName +": "+plain.body.replace(/@05[\da-f]{64}/g, '@hidden')));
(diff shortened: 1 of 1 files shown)


majestrate:
this doesn't really solve the underlying issue as you can still get session ids of posters via `GET /room/room-name-here/messages/recent` 
this is why we are doing key blinding, that code is already in the server side but clients do not currently implement blinded signatures.

ianmacd:
> this doesn't really solve the underlying issue as you can still get session ids of posters via `GET /room/room-name-here/messages/recent` this is why we are doing key blinding, that code is already in the server side but clients do not currently implement blinded signatures.

Ah, bummer. I didn't know it was that trivial to grab Session ids.

I was hoping this fix would tide us over until the Session clients have the blinding code merged.

jagerman:
Closing as this is addressed by blinding which is now released and on-by-default.

PR #105, opened 2022-05-12
Catch exception when blinding fails

jagerman:
If there is some invalid session_id it could result in sodium failing
the blinding calculation; in such a case just warn and skip it.


majestrate (commented):
i would either omit the query on only catch blinding operation failure exception type

jagerman:
> i would either omit the query on only catch blinding operation failure exception type

Fixed.

majestrate (approved):

PR #109, opened 2022-06-25
API docs generation using docsify

jagerman:
Generate nice API documentation using docsify.  Sample at https://api.oxen.io/sogs

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -1,3 +1,0 @@
-site_name: SOGS API
-theme: nature
-markdown_extensions: [codehilite]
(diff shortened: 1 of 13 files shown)


docs/mkdocs.yml
```diff
@@ -1,3 +0,0 @@
-site_name: SOGS API
-theme: nature
-markdown_extensions: [codehilite]
```
majestrate, line 3: maybe keep some notes around about generating static html without js from the markdown for those who want it.

jagerman: Then we'd effectively be supporting both doc generation systems.  I feel like someone who objects to javascript is probably perfectly happy just reading the markdown directly (or maybe even just the pydoc directly).

majestrate (approved):

PR #111, opened 2022-06-28
Upgrade test fixes

jagerman:
- Active users was being pruned and thus breaking the test.
- Remove warning from an unnecessary call to cleanup() (which can't be called as was being done here).


majestrate (approved):
rubberstamp

PR #113, opened 2022-06-30
Reactions

jagerman:
Adds support for server-side reactions handling for open groups.

Reactions are now returned as an extra field when messages are retrieved, such as this:
```json
        "reactions": {
          "👍": {
            "index": 1,
            "total": 150,
            "reactors": [
              "058328640343b91c03d393b8dc6ce15c8f93b191ff78054b6ae2c8030e2680b4d6",
              "05d782a81448199aea940f496cc70dc644c269e5979e472536ea03a067bf46d54a",
              "05233432db6edcd320a06b550d4dcf1a06cc486c58f8017a565f6f7dc0fc8b9512",
              "05ca519a397a1aa8a7a1515ff433dd2784cbcbb31c358b1a6e97fb758ceab44dca"
            ],
            "you": true
          },
          "👎": {
            "index": 0,
            "total": 27,
            "reactors": [
              "05ca519a397a1aa8a7a1515ff433dd2784cbcbb31c358b1a6e97fb758ceab44dca",
              "055297c3dc032bef33a21dcb1b76265c8319e53ed0867accd3c274c25ae33c1731",
              "05986c4e7fec0ac1a3a544d84367b2e995f85a6356be8d082724781a9ff699ac7a",
              "0537f1727302b70be32fc371523097bcdbc3e681f947a68ef5e3ea1365a689666a"
            ]
          },
          "🍆": {
            "index": 2,
            "total": 1,
            "reactors": [
              "051234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"
            ],
            "you": true
          },
```

When polling for new messages (using the `/since` endpoint) you also get updates for any posts that have had new reactions added or removed from the post (but not otherwise changed); these look vaguely similar to messages that you'd normally get back from `/since`, except they are missing most of the info other than id, seqno, and the reactions:
```json
{ "id" 1234, "seqno": 56768, "reactions": { /* as above */ } }
```
so that, when polling, you get notified (in bulk) of reaction changes in one go per message.

Since this might break existing clients, I've feature-guarded this by adding a new URL parameter `t=r` to the `/since` endpoint, so that you'd fetch: `/room/ROOMID/messages/since/56768?t=r` to tell SOGS that you want to opt-in to the new reactions rows.

Reactions update the room's sequence number, so clients should update their current seqno value when they get such a reaction row (so that they will only get the reactions again if there are changes to reactions).

The `reactors` shows the first N session IDs that reacted; by default this is 4, but can be anywhere from 0 to 20 by adding a `reactors=N` query string parameter.  (As a special case, 0 omits the "reactors" key entirely).

Adding/removing a reaction is done via endpoints:
Submitting reactions is done via:

`PUT /room/ROOM/reaction/MSGID/👍`  (body is an empty json dict `{}`)
`DELETE /room/ROOM/reaction/MSGID/👍`

(I want to say that the  👍 should be URL-encoded utf-8 bytes (e.g. `%f0%9f%91%8d` for the encoded 👍) when making a direct HTTP request (e.g. over lokinet), but from my testing uwsgi and sogs seem perfectly happy with a raw UTF-8 value there as well. Onion requests should be perfectly fine with raw utf8. I haven't yet tested whether it survives cleanly through an nginx proxy in the direct HTTP case yet).

Moderator endpoints are:

`DELETE /room/ROOMID/reactions/MSGID/🍆`

to delete all of a single reaction from a message, and:

`DELETE /room/ROOMID/reactions/MSGID`

to delete all reactions of any type from a message.

There is also an endpoint to get all reactors (not limited to the first N) of a message, including the timestamp when they reacted (this is probably not immediately useful for Session, but would be quite useful for, say, a bot that wants to check for all reactions on startup):

`GET /room/ROOMID/reactors/MSGID/🍆`

which returns lists of pairs of session ID + timestamp, sorted by timestamp, of when reactions were added.  For example:
```json
[
    ["05ca519a397a1aa8a7a1515ff433dd2784cbcbb31c358b1a6e97fb758ceab44dca", 1656631467.6396575],
    ["055297c3dc032bef33a21dcb1b76265c8319e53ed0867accd3c274c25ae33c1731", 1656631480.2537503]
]
```


sogs/model/room.py
```diff
@@ -529,16 +529,22 @@ def get_messages_for(
         recent: bool = False,
         single: Optional[int] = None,
         limit: int = 256,
+        reactions: bool = True,
+        reaction_updates: bool = True,
+        reactors: int = 0,
```
majestrate: `reaction_fetch_limit` instead of `reactors`?

sogs/model/room.py
```diff
@@ -979,6 +1042,207 @@ def attachments_size(self):
...
+    ):
+        """
+        Returns reaction information for the given messages.  Returns a dict of message id ->
+        reaction info for any messages in the given list that have reactions.  The reaction info is
+        itself a dict where keys are reaction strings and values are a dict containing keys:
+
+        - `"count"` -- the total number of this reaction
+        - `"you"` -- boolean set to true if `user` is given and the user has applied this reaction;
+          omitted otherwise.
+        - `"reactors"` -- if the input parameter `reactors` is greater than 0 then this is a list of
```
majestrate: would having reactions paginated make more sense vs just a limit?

sogs/model/room.py
```diff
@@ -979,6 +1042,207 @@ def attachments_size(self):
...
+        you = (
+            """
+            EXISTS(
+            SELECT * FROM reactions WHERE message = r.message AND reaction = r.reaction
+                AND "user" = :uid)
+            """
+            if user is not None
+            else 'FALSE'
+        )
+        for msgid, react, count, you in query(
```
majestrate: shadowing `you` here feels really jenk

sogs/model/room.py
```diff
@@ -979,6 +1042,207 @@ def attachments_size(self):
...
+                COUNT(*) AS react_count,
+                {you} AS you
+            FROM reactions r
+            WHERE message IN :msgs GROUP BY message, reaction
+            """,
+            bind_expanding=('msgs',),
+            msgs=message_ids,
+            uid=None if user is None else user.id,
+        ):
+            reacts.setdefault(msgid, {})[react] = (
```
majestrate: can reacts be a `collections.defaultdict` instead?

sogs/model/room.py
jagerman: Not really; the intended use case is that you get the reaction, and for each one, get the first N users who reacted to show a message such as

"Jon, 051234..., 05767195... and 30 more"

and if you *really* need the entire list (which clients usually won't, but a bot might), go fetch it with the dedicated reactors endpoint.

sogs/model/room.py
jagerman: That seems misleading because we aren't limiting how many reactions we fetch: we fetch all the reactions, and for each one, provide a limited list of the first users (i.e. "reactors") to react to it.

sogs/model/room.py
jagerman: Unintentional and fixed.

sogs/model/room.py
jagerman: It could, but it wouldn't save anything: this is the only place we use `setdefault` on it.

sogs/model/room.py
jagerman: Renamed to `reactor_limit`

sogs/model/room.py
jagerman: I don't know if this endpoint is even going to be used much; it's mainly there for debugging and, possibly, for a bot that needs to sync up reactions -- but I think even that latter case can be better addressed by something in the bot zmq api.

jagerman: Also: paginations suck for something like this because, typically, if you're using this endpoint (rather than just getting the first twenty from the message endpoint) you're going to get all of them and will end up doing a `LIMIT x OFFSET y`, and walk your way through all the pages, which just makes everyone slower (because you basically have to do the full database selection for each page, then throw away most of the results).

jagerman:
~~TODO:~~
- [x] apply reaction ordering table changes to db migration code

jagerman:
This also includes the Session-team requested `"deleted": true` flag for a message deletion update (equivalent to looking for `"data": null`, but might be easier to notice with some parsing mechanisms).

majestrate (approved):

PR #123, opened 2022-08-24
Room return info_updates

jagerman:
This works similarly to returning the seqno from post-modifying endpoints, but for room metadata (fields, pinned messages, mods, etc.).

This allows a client to track when the expected change should have been applied to the room when there might be concurrent room polling that could race with a pre-update response.

This PR also fixes a bug found in the added tests where the room's info_updates wasn't being updated when hidden global mods are added/removed, which would cause Session clients to not update the mods list for such a change.


majestrate (approved):
i cannot find anything wrong with this and ci passes

PR #128, opened 2022-09-25
Proof of concept: Reject postings containing text written in Farsi.

ianmacd:
In its current state, the filter is of limited use:

* The filter allows only a single group to be whitelisted from filtering. This makes it unusable if Farsi is an allowed language in more than one group on the server.

* No error other than the failure to send is reported to the client.

The purpose of this PR is to spark discussion and pave the way for the implementation of a multi-lingual, per group language filter.

--
Codepoint reference: https://github.com/mirhmousavi/Regex.Persian.Language

### DIFF
--- a/sogs.ini.sample
+++ b/sogs.ini.sample
@@ -115,6 +115,16 @@
 ;
 ;profanity_custom =
 
+; Whether we should reject messages written in Farsi.
+;
+;language_filter_farsi = no
+
+
+; If we should reject messages written in Farsi, we should still allow them in the
+; room with this id. 0 means reject from all rooms.
+;
+;language_whitelist_farsi = 0
+
 
 [web]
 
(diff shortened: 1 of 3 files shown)


ianmacd:
The filter can now be enabled with `language_filter_farsi = yes` in `sogs.ini`, and a group id can be specified to exempt it from filtering, for example `language_whitelist_farsi = 18`.

Obviously it would be much better to allow a list of group ids here.

ianmacd:
I have now totally overhauled this in #129, which supersedes this pull-request.

#129 more accurately addresses the issue of alphabet filtering, since that's what this PR was actually doing.

The new PR also allows for multiple alphabet filtering and per-alphabet whitelisting.

jagerman:
Closing (replaced with #129).

PR #129, opened 2022-09-26
Allow rejection of postings written in a particular alphabet.

ianmacd:
The user configures in `sogs.ini` a list of alphabets to reject by default:

`alphabet_filters = [ arabic, cyrillic, persian ]`

An empty list means no filtering is desired.

If there are rooms in which these alphabets should be allowed, a list of room ids per alphabet may be specified for whitelisting.

`alphabet_whitelist_arabic = [ 8 ]`
`alphabet_whitelist_cyrillic = [ 5, 13 ]`
`alphabet_whitelist_persian = [ 18 ]`

A list of valid room ids can be obtained with the following SQL:

`$ sudo sqlite3 /var/lib/session-open-group-server/sogs.db 'SELECT * FROM rooms;'`

The implementation can easily be extended to more alphabets.

This pull-request supersedes #128.

### DIFF
--- a/sogs.ini.sample
+++ b/sogs.ini.sample
@@ -115,6 +115,20 @@
 ;
 ;profanity_custom =
 
+; Whether we should reject messages that use a particular alphabet.
+;
+;alphabet_filters = [ arabic, cyrillic, persian ]
+;alphabet_filters = []
+
+
+; If we reject messages written in a given alphabet, we should still allow them in
+; specific rooms. A list of whitelisted room ids can be given here, as returned by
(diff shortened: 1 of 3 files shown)


ianmacd:
Multiple rooms per alphabet can now be whitelisted.

jagerman:
I took this PR and ran with it in PR #131, most notably to support sending a whispered (or public, if you want) response to the user whose message is dropped (but also reworked how the config settings work).

jagerman:
Closing; this functionality is now included in #131

PR #138, opened 2022-10-18
Add test for extra deleted message reaction updates

jagerman:
There were some reports that this pattern could result in extra reaction updates, but that doesn't appear to be the case.  Commit the test anyway.


jagerman:
Closing; this wasn't a sogs issue after all (the issue it triggered in session-desktop is addressed by #143 )

PR #152, opened 2022-11-22
delete dm endpoints

### LINKED ISSUE #151: add endpoint for DM deletion
add endpoints for a user to delete their DMs.

majestrate:
fixes #151

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,3 +21,5 @@ install_requires=
     oxenmq
     oxenc
     pyonionreq
+    tomli
+    sqlalchemy
--- a/sogs/model/message.py
+++ b/sogs/model/message.py
@@ -47,19 +47,12 @@ class Message:
         """Delete all messages sent to a user or from a user.
         returns the number of rows affected.
         """
-        if sender and recip:
-            raise ValueError("delete_all(): cannot provide both sender and recipient")
+        if sum(bool(x) for x in (sender, recip)) != 1:
+            raise ValueError("delete_all(): exactly one of sender or recipient is required")
 
-        recip_id = recip.id if recip else None
-        sender_id = sender.id if sender else None
         result = query(
-            f"""
-            DELETE FROM inbox
-            {'WHERE recipient = :recip' if recip else '' }
-            {'WHERE sender = :sender' if sender else '' }
-            """,
-            recip=recip_id,
-            sender=sender_id,
+            f"DELETE FROM inbox WHERE {'recipient' if recip else 'sender'} = :id",
+            id=recip.id if recip else sender.id,
         )
         return result.rowcount
 
--- a/sogs/model/message.py
+++ b/sogs/model/message.py
@@ -43,6 +43,27 @@ class Message:
         self._row = row
 
     @staticmethod
+    def delete_all(*, recip=None, sender=None):
+        """Delete all messages sent to a user or from a user.
+        returns the number of rows affected.
+        """
+        if sender and recip:
+            raise ValueError("delete_all(): cannot provide both sender and recipient")
+
+        recip_id = recip.id if recip else None
+        sender_id = sender.id if sender else None
+        result = query(
+            f"""
+            DELETE FROM inbox
+            {'WHERE recipient = :recip' if recip else '' }
+            {'WHERE sender = :sender' if sender else '' }
+            """,
+            recip=recip_id,
+            sender=sender_id,
+        )
+        return result.rowcount
+
+    @staticmethod
     def to(user, since=None, limit=None):
         """get all message for a user, returns a generator"""
         rows = query(
--- a/sogs/routes/auth.py
+++ b/sogs/routes/auth.py
@@ -121,6 +121,27 @@ def user_required(f):
     return required_user_wrapper
 
 
+def require_blind_user():
+    """Requires that the authenticated user is using a blinded pubkey for auth; aborts with 401
+    Unauthorized if the user has not authenticated with a blinded pubkey."""
+    require_user()
+    if not g.user.is_blinded:
+        abort_with_reason(http.UNAUTHORIZED, "This endpoint requires blinded pubkeys be used")
+
+
+def blind_user_required(f):
+    """Decorator for an endpoint that requires a user that is using a blinded public key;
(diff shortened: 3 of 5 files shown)


setup.cfg
```diff
@@ -21,3 +21,5 @@ install_requires=
     oxenmq
     oxenc
     pyonionreq
+    tomil
```
jagerman: What is this?  I can't find any such python package.

majestrate: typo, it is tomli

sogs/http.py
```diff
@@ -1,6 +1,7 @@
 # success codes:
 OK = 200
 CREATED = 201
+NO_CONTENT = 204
```
jagerman: I don't think we should use this; everywhere else in SOGS where we have nothing to return we're still returning application/json content of an empty dict (`{}`) with 200 OK, which allows us flexibility to add return values later without changing the status code.

sogs/routes/auth.py
```diff
@@ -121,6 +121,27 @@ def required_user_wrapper(*args, **kwargs):
     return required_user_wrapper
 
 
+def require_blind_user():
+    """Requires that the authenticated user is using a blinded pubkey for auth; aborts with 401
+    Unauthorized if the user has not authenticated with a blinded pubkey."""
+    require_user()
+    if not g.user.is_blinded:
+        abort_with_reason(http.FORBINDDEN, "This endpoint requires blinded pubkeys be used")
```
jagerman: Typo: `FORBIDDEN`

sogs/model/message.py
```diff
@@ -42,6 +42,25 @@ def __init__(self, row=None, *, sender=None, recip=None, data=None):
         assert row is not None
         self._row = row
 
+    @staticmethod
+    def delete_all(*, recip=None, sender=None):
+        """Delete all messages sent to a user or from a user.
+        returns the number of rows affected.
+        """
+        assert recip or sender
```
jagerman: This should throw an exception rather than a debug assert.  But also, given the query construction below, it needs to be the case that the caller has specified exactly one of these, rather than at least one: otherwise this function is going to generate an invalid query.

sogs/routes/dm.py
```diff
@@ -112,3 +110,25 @@ def send_inbox(sid):
...
+
+    204 No Content - On successful removal of messages.
+    304 Not Modified - If there were no messages to remove.
+
+    """
+
+    num_deleted = 0
+    with db.transaction():
+        num_deleted += Message.delete_all(recip=g.user)
+        num_deleted += Message.delete_all(sender=g.user)
```
jagerman: I'm don't think we should allow deleting sent messages.  Sent messages are the property of the recipient in regular Session message model (at best you can send another deletion meta-request follow-up) and I'm thinking that should be mirrored here.

sogs/routes/dm.py
```diff
@@ -112,3 +110,25 @@ def send_inbox(sid):
...
+
+@dm.delete("/inbox")
+@auth.blind_user_required
+def delete_inbox_items():
+    """
+    Deletes the user's sent and recieved messages.
+    # Return value
+
+    204 No Content - On successful removal of messages.
+    304 Not Modified - If there were no messages to remove.
```
jagerman: I don't like either of these statuses:

- On success everywhere else in SOGS we return a 200 OK with a json result (even if empty).

- 304 is a redirection response code for cache-control headers which which seems deeply inappropriate for a DELETE request response.  A 200 response code seems appropriate in this case: we asked to delete all, and it succeeded; the fact that there were none to delete doesn't mean it didn't have the intended effect.

sogs/routes/dm.py
```diff
@@ -112,3 +110,25 @@ def send_inbox(sid):
...
+    """
+
+    num_deleted = 0
+    with db.transaction():
+        num_deleted += Message.delete_all(recip=g.user)
+        num_deleted += Message.delete_all(sender=g.user)
+
+    if num_deleted:
+        return b'', http.NO_CONTENT
+    return b'', http.NOT_MODIFIED
```
jagerman: Since we have it, we might as well include the number of deleted messages in the response.

sogs/routes/dm.py
jagerman: Additionally, this request is for `DELETE /inbox`, so deleting everything in the outbox as well doesn't seem right.

(But as to my previous point: I don't think we should have a `DELETE /outbox`).

sogs/model/message.py
```diff
@@ -42,6 +42,25 @@ def __init__(self, row=None, *, sender=None, recip=None, data=None):
...
+        recip_id = recip.id if recip else None
+        sender_id = sender.id if sender else None
+        result = query(
+            f"""
+            DELETE FROM inbox WHERE
+            {'recipient = :recip' if recip else '' }
+            {'sender = :sender' if sender else '' }
+            """,
```
jagerman: This query is going to be invalid if you pass in both a `recip` and ` sender` at the same time.

sogs/routes/auth.py
```diff
@@ -121,6 +121,27 @@ def required_user_wrapper(*args, **kwargs):
     return required_user_wrapper
 
 
+def require_blind_user():
+    """Requires that the authenticated user is using a blinded pubkey for auth; aborts with 401
+    Unauthorized if the user has not authenticated with a blinded pubkey."""
```
jagerman, line 126: Status code in the docstring doesn't match the code.

jagerman (changes requested):

sogs/model/message.py
```diff
@@ -42,6 +42,27 @@ def __init__(self, row=None, *, sender=None, recip=None, data=None):
         assert row is not None
         self._row = row
 
+    @staticmethod
+    def delete_all(*, recip=None, sender=None):
+        """Delete all messages sent to a user or from a user.
+        returns the number of rows affected.
+        """
+        if sender and recip:
+            raise ValueError("delete_all(): cannot provide both sender and recipient")
```
jagerman: Also needs to fail if you pass in neither (because otherwise this query is going to delete all DMs of all users if two `None`s get passed in here).

jagerman (approved):

PR #165, opened 2023-01-09
Better room modification CLI arguments; allow --description modification

jagerman:
Room modification arguments were sort of a shambles: there were various "action" items like "--set-perms" or "--add-moderator", but those options were exclusive (meaning you couldn't, say, add a moderator and delete another moderator in the same command).

This overhauls how modifications work, accomplishing the following:
- removed the `--set-perms` argument; it is implied by having any of the `--add-perms`/`--remove-perms`/`--clear-perms` options.
- `--description` can now be used to update a room description (when used with `--rooms` to specify the room).  Previously it was only used when adding a room.
- Modifying options (like `--description`, `--add-moderators`, etc.) are no longer mutually exlusive: one command can now add/delete moderators, set permissions, and change the description all in one go.
- Simplified various identical logic around how `--rooms` is handled.

### DIFF
--- a/sogs/__main__.py
+++ b/sogs/__main__.py
@@ -192,7 +192,9 @@ if update_room and not args.rooms:
     )
     sys.exit(1)
 if args.rooms and not update_room:
-    print("Error: --rooms is not compatible with that option", file=sys.stderr)
+    # If we have --rooms but didn't recognize any of the `update_rooms` options then that means
+    # `--rooms` was specify with some action (e.g. `--initialize`) that doesn't support --rooms:
+    print("Error: --rooms specified without a room modification option", file=sys.stderr)
     sys.exit(1)
 
 from . import config, crypto, db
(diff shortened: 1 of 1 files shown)


sogs/__main__.py
```diff
@@ -147,6 +161,40 @@
...
+]
+for i in range(1, len(incompat)):
+    for j in range(0, i):
+        if incompat[j][1] and incompat[i][1]:
+            print(f"Error: {incompat[j][0]} and {incompat[i][0]} are incompatible", file=sys.stderr)
+            sys.exit(1)
+
+if update_room and not args.rooms:
+    print(
+        "A room must be specified (using --rooms) when updating permissions or room details",
```
tewinget, line 193:
```suggestion
        "A room must be specified (using --rooms) when updating room permissions or details",
```

tewinget: nevermind, misunderstood

tewinget (approved):

PR #167, opened 2023-01-19
Add `/rooms/all/<SessionID:sid>` for server-wide message deletion.

ianmacd:
This causes all messages by the given Session id to be deleted from the server. The deleting user must be a global moderator on the server.

This facility is needed to allow a server-wide 'Ban and Delete' feature for Session moderators and admins.

### DIFF
--- a/sogs/routes/rooms.py
+++ b/sogs/routes/rooms.py
@@ -960,3 +960,33 @@ def delete_all_posts(room, sid):
     if not deleted:
         abort(http.NOT_FOUND)
     return jsonify({})
+
+
+@rooms.delete("/rooms/all/<SessionID:sid>")
+def delete_user_posts_from_all_rooms(sid):
+    """
+    Deletes all posts from all rooms by a given user.
+
+    # URL Parameters
+
+    - `sid` — the session id of the user to ban
+
+    # Return value
+
+    A JSON dict with the keys:
+
+    - `total` — The total number of posts deleted across all rooms.
+    - `rooms` — A dict of room tokens and their deletion counts.
+    """
+    deletions = {}
+    total = 0
+    user = muser.User(session_id=sid, autovivify=False)
+    for room in mroom.get_accessible_rooms(g.user):
+        try:
+            count, _ = room.delete_all_posts(user, deleter=g.user)
+            total += count
+            deletions[room.token] = count
+        except exc.BadPermission:
+            pass
+
+    return jsonify({"total": total, "rooms": deletions})
--- a/sogs/routes/users.py
+++ b/sogs/routes/users.py
@@ -293,7 +293,7 @@ def ban_user(sid):
 
     The user's messages are not deleted by this request.  In order to ban and delete all messages
     use the [`/sequence`](#post-sequence) endpoint to bundle a `/user/.../ban` with a
-    [`/user/.../deleteMessages`](#post-usersiddeleteMessages) request.
+    [`/rooms/all/...`](#delete-roomsallsid) request.
 
     # Return value
 


sogs/routes/rooms.py
```diff
@@ -960,3 +960,27 @@ def delete_all_posts(room, sid):
...
+
+    # Error status codes
+
+    - 403 Forbidden — if the invoking user does not have access to a room.
+    """
+    user = muser.User(session_id=sid, autovivify=False)
+    for room in mroom.get_accessible_rooms(g.user):
+        deleted, _ = room.delete_all_posts(user, deleter=g.user)
+        if not deleted:
+            app.logger.warning(f"Error deleting posts from room: {room}.")
```
jagerman: `deleted = 0` isn't really a failure case; it can just mean the user had no posts in that room, which would happen pretty often (since it's rare that someone spams in *every single* room).

Rather this call throws for permission errors (e.g. if you aren't a moderator of the room, or aren't an admin and are trying to delete an admin's posts), but in such a case we likely want to continue trying with other rooms, so something like:
```Python
for room in mroom.get_accessible_rooms(g.user):
    try:
        room.delete_all_posts(user, deleter=g.user)
    except BadPermission:
        pass
```

sogs/routes/rooms.py
```diff
@@ -960,3 +960,27 @@ def delete_all_posts(room, sid):
...
+    return jsonify({})
```
jagerman: It would be nice to return a summary of deletion counts here so that a client could display the summary; perhaps something like:
```Python
deletions = {}
total = 0
for room in ...
    try:
        count, _ = room.delete_all_posts(...
        total += count
        deletions[room.token] = count
    # ...

return jsonify({"total": total, "rooms": deletions})

File: sogs/routes/rooms.py [hunk 3]
```diff
@@ -960,3 +960,27 @@ def delete_all_posts(room, sid):
     if not deleted:
         abort(http.NOT_FOUND)
     return jsonify({})
+
+@rooms.delete("/room/all/<SessionID:sid>")
```
jagerman: `/room/ANYTHING/...` is reserved for specific access into a single room (so that if someone wants to create a room with id "all" it wouldn't conflict with the endpoint added here).  There is, however, the `/rooms` prefix already that we could add onto (currently it only has the top-level target to return a list of all server rooms) by renaming this endpoint to `/rooms/all/<SessionID:sid>`

jagerman: Also: the non-existent `/user/.../deleteMessages` endpoint mentioned in routes/users.py can get updated to match this new endpoint.

ianmacd: Thanks. Changed as described.

sogs/routes/rooms.py
ianmacd: Much better. Thanks.

sogs/routes/rooms.py
ianmacd: A great example of how to take a simple snippet of code and make it so much better.

Thank you for all these suggestions.

Changed as described.

sogs/routes/rooms.py
```diff
@@ -960,3 +960,37 @@ def delete_all_posts(room, sid):
...
+    # Return value
+
+    A JSON dict with the keys:
+
+    - `total` — The total number of posts deleted across all rooms.
+    - `rooms` — A dict of room tokens and their deletion counts.
+
+    # Error status codes
+
+    - 403 Forbidden — if the invoking user does not have access to a room.
```
jagerman: I don't think this error can happen from here anymore; I think if you don't have any access you'll just get 0 deletions.

jagerman (approved):

sogs/routes/rooms.py
```diff
@@ -960,3 +960,37 @@ def delete_all_posts(room, sid):
...
+    """
+    deletions = {}
+    total = 0
+    user = muser.User(session_id=sid, autovivify=False)
+    for room in mroom.get_accessible_rooms(g.user):
+        try:
+            count, _ = room.delete_all_posts(user, deleter=g.user)
+            total += count
+            deletions[room.token] = count
+        except BadPermission:
```
jagerman: Oops, my mistake, that should be `exc.BadPermission`

sogs/routes/rooms.py
ianmacd: Yeah, I wondered myself about that.

Removed.

sogs/routes/rooms.py
ianmacd: Fixed.

sogs/routes/rooms.py
jagerman: Looks like this only got half-removed.

ianmacd: A good thing one of us is paying attention.

jagerman (approved):

jagerman:
Great, thanks for the PR!

PR #174, opened 2023-02-09
Bot API Implementation

dr7ana:
## Draft PR

Ideal Control Flow:
1) message comes in HTTP request
2) unpacked/parsed/verified/permissions checked
3) comes into relevant route (ex: add_post())
4) sends off to mule to be handled by bots
5) mule has ordered list of bots by priority
6) mule passes message to bots, which have fixed return values (insert, do not insert)
7) if all bots approve, mule replies to worker with go ahead or vice versa for no go

Architectural Changes:
- OMQ object is siloed in [OMQ class](https://github.com/dr7ana/session-pysogs/blob/cbaaecef762981e77552afe6312ad7d42e3ff48e/sogs/omq.py#L13)
- Mule object is siloed in [Mule class](https://github.com/dr7ana/session-pysogs/blob/d8836d5f67668b6ef0c47f8ba7758300a9388070/sogs/mule.py#L32) and is accessible by global variable [omq_global](https://github.com/dr7ana/session-pysogs/blob/cbaaecef762981e77552afe6312ad7d42e3ff48e/sogs/routes/messages.py#L364)
- Bots are managed by [Manager class](https://github.com/dr7ana/session-pysogs/blob/d8836d5f67668b6ef0c47f8ba7758300a9388070/sogs/model/manager.py#L74) and held in a self-ordering "priority-queue"-like dictionary
- User permission/authorization functionalities siloed in sogs/authentication/authority.py

TODO:
- Create root-level folder for all authentication things related to user permissions; bot manager should just check for things like `is_allowed_to_message()`
- Add OMQ-level logic to adding and removing bots
- Flesh out a sample bot implementation (to replace model/bot.py) with SimpleFilter functionality to test against
  - Implement iterative polling logic in BotQueue 
  - Test edge cases in bot polling


sogs/mule.py
```diff
@@ -86,18 +29,78 @@ def wrapper(*args, **kwargs):
...
+        worker = self._omq.add_category("worker", access_level=oxenmq.AuthLevel.admin)
+        worker.add_command("message_posted", self.message_posted)
+        worker.add_command("messages_deleted", self.messages_deleted)
+        worker.add_command("message_edited", self.message_edited)
+
+        ## NEW CODE FOR BOT
+        handler = self._omq.add_category("handler", access_level=oxenmq.AuthLevel.admin)
+        handler.add_command("add_bot", omq.add_bot)
+        handler.add_command("remove_bot", omq.remove_bot)
+        handler.add_command("send_to_handler", omq.manager.receive_message)
```
darcys22: Unsure if admin will be the right for sending to handler. Access level potentially different for a bot being added from an external connection with admin vs internal message to mule.

sogs/model/bot.py
```diff
@@ -0,0 +1,267 @@
...
+        self.rooms.remove(_room)
+
+        if not self.rooms:
+            delete_bot_function_goes_here = True
+
+    def _refresh_cache(self):
+        for r in self.rooms:
+            self.perm_cache = self.perm_cache | r._perm_cache
+
+    def check_permission_for(
```
darcys22: Think this needs to be separated out, Authentication is a pretty important thing in its own right and SOGS should handle that outside of the Bot.

darcys22: Also looks to be a duplicated here and Manager

sogs/model/room.py
```diff
@@ -1151,7 +764,7 @@ def delete_posts(self, message_ids: List[int], deleter: User):
                 )
 
                 if ids:
-                    if not self.check_moderator(deleter):
+                    if not self.default_bot.check_moderator(self, deleter):
```
darcys22, line 765: These `default_bot.check_moderator` and other similar functions feel wrong. 

The call should be something more like `user.is_moderator` and have an authentication module in the root level directory that fills out this when the request is received

sogs/model/bot.py
```diff
@@ -0,0 +1,267 @@
...
+from .room import Room
+from .message import Message
+from .filter import SimpleFilter
+from .exc import InvalidData
+
+from typing import Optional, List, Union
+import time
+
+
+class Bot:
```
darcys22: I'd probably drag this whole file out of here and put it in its own `utils/bots/filter_bot.py` file. If its detached completely from the core then users will be able to copy and paste this Bot and make adjustments to turn it into whatever they want.

The bot itself wont want to touch anything like the database or models because its purely working though an OMQ connection and whatever API we define so the bot can communicate with the mule.

sogs/model/filter.py
```diff
@@ -0,0 +1,212 @@
...
+
+from .. import crypto, db, config
+from ..db import query
+from ..hashing import blake2b
+from ..web import app
+from nacl.signing import SigningKey
+from .exc import NoSuchUser, BadPermission, PostRejected
+from sogs.model.user import User
+from sogs.model.bot import Bot
+from sogs.model.room import Room, alphabet_filter_patterns
```
darcys22, line 11: Should try get this Filter to exist separately from these Models.

Like it just gets fed a raw text message checks if its naughty according to its configured filters. Then returns a bool if its safe. 

That way this can be extracted out and used by the bots when the bots themselves have no access to the models

sogs/model/filter.py
```diff
@@ -0,0 +1,212 @@
...
+                    )
+                )
+                filter_privkeys[prof_name] = signingkey
+
+            sig = signingkey.sign(pbmsg).signature
+            server_fake_user = User(
+                session_id='15' + signingkey.verify_key.encode().hex(), autovivify=True, touch=False
+            )
+
+            def insert_reply():
```
darcys22, line 180: Yeah the filter shouldnt insert into the database.

sogs/model/manager.py
```diff
@@ -0,0 +1,300 @@
...
+        id: Optional[int] = None,
+        session_id: Optional[int] = None,
+    ) -> None:
+        self.id = id
+        self.queue = BotQueue()
+
+    def qempty(self):
+        return not self.queue._empty()
+
+    def add_bot(self, bot: Bot, priority: int = None):
```
darcys22: A `Bot` here would be a connection to a bot rather than a Class that can be executed.

sogs/model/room.py
```diff
@@ -71,14 +67,25 @@ class Room:
...
         default_write - True if default user permissions includes write permission
         default_upload - True if default user permissions includes file upload permission
+
+        NEW:
+        default_bot - default_bot object reference
+        active_bot - true if default_bot moderator is actively moderating or false if passively filtering
     """
 
-    def __init__(self, row=None, *, id=None, token=None):
+    def __init__(self, row=None, *, id=None, token=None, bot: Bot = None):
```
darcys22, line 76: The room model shouldnt need to know what bots exist. It should just write and read and let the manager handle the bot communications before it even gets to here

sogs/mule.py
dr7ana: yeah that's a good point. will likely need a new category for that, as "add_bot" and "remove_bot" make sense to me as admin level tasks. thoughts?

sogs/model/bot.py
dr7ana: agreed RE: authentication. this file is leftover from first draft, i'm converting it today to a class representing a link to a bot

sogs/model/room.py
dr7ana: oh god, this is the last time i use VS code's "refactor" feature. it replaced every instance of bot with self.default_bot.......... will fix

sogs/model/filter.py
dr7ana: agreed. we can turn it into a basic bot to use for testing

sogs/model/bot.py
```diff
@@ -0,0 +1,267 @@
...
+        self.rooms: List[Room] = _rooms
+        self.status = {r: False for r in self.rooms}
+        self.filter = SimpleFilter(_bot=self)
+        self.perm_cache = {}
+        self.current_message: Message = None
+        self.nlp_model = None
+        self.language = 'English'
+        self.word_blacklist = ['placeholderlist', 'until', 'we', 'decide', 'naughty', 'words']
+
+    def __setattr__(self, __name: str, __value) -> None:
```
majestrate: why are we hooking `__setattr__` and `__getattr__` here?

the norm is usually that all `_` prefixed members are considered invisible to other types.

sogs/model/filter.py
```diff
@@ -0,0 +1,212 @@
...
+                if rf is not None:
+                    reply_format = random.choice(rf)
+                if pn is not None:
+                    profile_name = pn
+                if pb is not None:
+                    public = pb
+
+        return (reply_format, profile_name, public)
+
+    def read_message(self, user: User, data: bytes, room: Room):
```
majestrate, line 92: naming nitpick: `view_message` or `inspect_message` feel a bit less ambiguous. (does this read the message or does it tell us if we read the message?)

sogs/model/room.py
majestrate: is there a reason that room model can't have the `check_moderator` method that just calls `default_bot.check_moderator` ?

majestrate: (a lot less would have been modified that way)

jagerman: > oh god, this is the last time i use VS code.

Fixed that for you, I hope.

dr7ana: all removed

sogs/omq.py
```diff
@@ -6,62 +6,63 @@
...
+        be prefixed with "worker." (unless overridden).
 
-    omq.start()
-    app.logger.debug("Started, connecting to mule")
-    mule_conn = omq.connect_remote(oxenmq.Address(config.OMQ_INTERNAL))
+        Any args will be bt-serialized and send as message parts.
+        """
+        
+        if prefix:
+            command = prefix + command
```
majestrate: i feel like `prefix` should not require a trailing `.` to prevent footcannons.
then we can have `command = f"{prefix}.{command}"` here

sogs/model/bot.py
dr7ana: this class is not going to exist soon

sogs/model/filter.py
dr7ana: this class is holding all the previous filter functionalities that we can turn into a bot to test the interface

sogs/omq.py
dr7ana: agreed. fixed

sogs/model/room.py
```diff
@@ -17,21 +17,15 @@
...
 import re
 import sqlalchemy.exc
 import time
 from typing import Optional, Union, List
 
 
-# TODO: These really should be room properties, not random global constants (these
-# are carried over from old SOGS).
-rate_limit_size = 5
-rate_limit_interval = 16.0
```
jagerman, line 32: The TODO was poorly worded; it didn't mean properties as in object properties, but rather configurable attributes stored in the room table that an admin could set/update/etc.

PR #188, opened 2023-04-20
Increase max attachment size to 10MB

mdPlusPlus:
Untested, but should be in line with the recent client changes to allow for up to 10MB.


majestrate:
this is changing the default value, why do you want to change the default value?

mdPlusPlus:
Why should the default value be less than what's currently allowed?  
Sorry, am I misunderstanding something here?

jagerman:
Looks good to me.

The 6MB default limit was originally imposed because there were multiple inefficiencies passing binary data through onion requests, and so was matching a Session-side limit on file uploads that existed at the time so as to not exceed the maximum raw message size passable through onion requests (10MiB), but Session migrated to binary-efficient onion requests (aka "v4 onion requests") quite a while ago, and recently also updated the maximum attachment size in DMs to 10MB (MB, not MiB), so it makes sense to update this to match.

(I also just did a test upload also to make sure this works properly for upload and download of a 9.99MB file, and it did).

jagerman (approved):

PR #191, opened 2023-06-13
Fix the nginx config proxy_pass pattern

Pingmin:
Notes:

When use the ` http://127.3.2.1:4242$request_uri ` in the nginx config file `sogs-proxy`, it will redirect to a ERROR page ` http://127.3.2.1:4242/r/room_token ` by clicking the room link on the SOGS index and rooms web page.

When use the ` http://127.3.2.1:4242/ ` instead, it will redirect to ` http://example.com/r/room_token/ ` or ` http://<public_ip>/r/room_token `, and this works well for the SOGS index web accessing.

I just found, fixed and tested this recently!

### DIFF
--- a/debian/sogs-proxy.nginx.conf
+++ b/debian/sogs-proxy.nginx.conf
@@ -9,7 +9,7 @@ server {
     }
     location / {
         proxy_set_header X-Real-IP $remote_addr;
-        proxy_pass http://127.3.2.1:4242$request_uri;
+        proxy_pass http://127.3.2.1:4242/;
     }
 
     listen 80;


jagerman:
> I just found, fixed and tested this issue recently!

This looks good to me.  Have you confirmed that this works for the whole sogs (i.e. not just the index web pages)?

Pingmin:
What does the "whole sogs" mean? 

I just tested both the top index page and the sub room's pages, and used the latest Session Desktop on Ubuntu 22.04 to send messages.

jagerman:
> and used the latest Session Desktop on Ubuntu 22.04 to send messages.

Yeah, this is what I meant (i.e. Session clients, not just the web UI).

jagerman:
Added (outside github) in 6d9362b

jagerman:
Thanks for the fix; I've upload 0.3.7-2 with this applied to our repos.

PR #194, opened 2023-06-19
Update administration.md

ChillyKitty:
Added more administration guidance and information.

### DIFF
--- a/administration.md
+++ b/administration.md
@@ -111,23 +111,17 @@ sogs -L
 
 This includes details such as the number of messages, files, active users, and moderators.  If you
 also want to list each of the individual moderators in each room add `-v` to the end of the command.
-## List room ids
-With sqlite3 installed run
-```
-sqlite3 /var/lib/session-open-group-server/sogs.db 'select id,token from rooms;'
-```
-## Clear out old room messages
-After getting the room id you can clear out messages stored on your server since this is not done automatically. In the example below our room id is 1 and messages older than 14 days are being deleted. 
-```
-sqlite3 /var/lib/session-open-group-server/sogs.db "DELETE FROM message_details WHERE room = 1 AND posted < strftime('%s', 'now', '-14 days');"
-```
+
 ## Backup sogs
 It is recommended to do periodic backups of your sogs instance. You can backup the needed files to your current directory by running
 ```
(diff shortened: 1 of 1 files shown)


administration.md
```diff
@@ -111,7 +111,26 @@ sogs -L
...
+## List room ids
+With sqlite3 installed run
+```
+sqlite3 /var/lib/session-open-group-server/sogs.db 'select id,token from rooms;'
+```
+## Clear out old room messages
+After getting the room id you can clear out messages stored on your server since this is not done automatically. In the example below our room id is 1 and messages older than 14 days are being deleted. 
+```
+sqlite3 /var/lib/session-open-group-server/sogs.db "DELETE FROM message_details WHERE room = 1 AND posted < strftime('%s', 'now', '-14 days');"
+```
```
jagerman: A better query here would be this so that you don't have to worry about the ID values at all:
```sql
DELETE FROM message_details WHERE room = (SELECT id FROM rooms WHERE token = 'someroom') AND posted < strftime('%s', 'now', '-14 days');
```

However, in general, I'd prefer that we moved such functionality into the `sogs` command-line tool as I'm rather wary of having people directly interact with the sqlite database.

administration.md
```diff
@@ -111,7 +111,26 @@ sogs -L
...
+## Backup sogs
+It is recommended to do periodic backups of your sogs instance. You can backup the needed files to your current directory by running
+```
+#stop sogs server so files can be backed up (assumes you you systemd service for sogs)
+systemctl stop sogs;
+#backup files to tar.xz in current directory with current date.
+tar -cJvf pysogs_backup_$(date --iso-8601).tar.xz /var/lib/session-open-group-server /etc/sogs/sogs.ini;
+#starts sogs (assumes you you systemd service for sogs)
+systemctl start sogs;
```
jagerman, line 126: :+1: This looks good to me

jagerman:
Backup for a SOGS using postgresql instead of SQLite would need a separate dump of the database via:

```
pg_dump -Fc sogs >pysogs_db_backup_$(data --iso-8601).pgdump
```

alongside the tar file from the sqlite instructions.

administration.md
ChillyKitty: I agree, was saying this in the Session SOGS group. I'll look into it.

